PR #13, opened 2018-05-17
avoid allocations when unmarshaling

nussjustin:
Currently `readAllAppend` allocates a `*bytes.Buffer` on each call. Also for slices with less than bytes.MinRead bytes (512) the buffer will allocate a new `[]byte` even if the old slice could fit all the data.

In most cases the capacity of the `[]byte` is rather small (64 bytes in most cases, which is the capacity used by `getBytes` for new slices), which means that most calls will allocate at least 3 times.

Avoid this by reading into pooled buffers and appending the data to the given `[]byte`.

Also add a benchmark that calls `readAllAppend` with different capacities to show the different allocation cases.

Benchmark results:

```
name                         old time/op    new time/op    delta
ReadAllAppend/Capacity0-8       965ns ± 3%      99ns ± 3%   -89.76%  (p=0.000 n=15+15)
ReadAllAppend/Capacity11-8      439ns ± 5%      66ns ± 2%   -85.02%  (p=0.000 n=15+13)
ReadAllAppend/Capacity523-8     121ns ± 6%      67ns ± 3%   -44.45%  (p=0.000 n=15+15)

name                         old alloc/op   new alloc/op   delta
ReadAllAppend/Capacity0-8      2.16kB ± 0%    0.02kB ± 0%   -99.26%  (p=0.000 n=15+15)
ReadAllAppend/Capacity11-8       688B ± 0%        0B       -100.00%  (p=0.000 n=15+15)
ReadAllAppend/Capacity523-8      112B ± 0%        0B       -100.00%  (p=0.000 n=15+15)

name                         old allocs/op  new allocs/op  delta
ReadAllAppend/Capacity0-8        3.00 ± 0%      1.00 ± 0%   -66.67%  (p=0.000 n=15+15)
ReadAllAppend/Capacity11-8       2.00 ± 0%      0.00       -100.00%  (p=0.000 n=15+15)
ReadAllAppend/Capacity523-8      1.00 ± 0%      0.00       -100.00%  (p=0.000 n=15+15)
```

### DIFF
--- a/resp/resp.go
+++ b/resp/resp.go
@@ -78,6 +78,22 @@ func putBytes(b *[]byte) {
 	bytePool.Put(b)
 }
 
+var bufferPool = sync.Pool{
+	New: func() interface{} {
+		b := make([]byte, 0, bytes.MinRead)
+		return bytes.NewBuffer(b)
+	},
+}
+
+func getBuffer() *bytes.Buffer {
+	return bufferPool.Get().(*bytes.Buffer)
+}
+
+func putBuffer(b *bytes.Buffer) {
+	b.Reset()
+	bufferPool.Put(b)
+}
+
 // parseInt is a specialized version of strconv.ParseInt that parses a base-10 encoded signed integer.
 func parseInt(b []byte) (int64, error) {
 	if len(b) == 0 {
@@ -179,12 +195,17 @@ func bufferedIntDelim(br *bufio.Reader) (int64, error) {
 }
 
 func readAllAppend(r io.Reader, b []byte) ([]byte, error) {
-	buf := bytes.NewBuffer(b)
+	buf := getBuffer()
 	// a side effect of this is that the given b will be re-allocated if
 	// it's less than bytes.MinRead. Since this b could be all the way from the
 	// user we can't guarantee it within the library
 	_, err := buf.ReadFrom(r)
-	return buf.Bytes(), err
+	b = append(b, buf.Bytes()...)
+	putBuffer(buf)
+	if b == nil {
+		b = []byte{}
+	}
+	return b, err
 }
 
 func multiWrite(w io.Writer, bb ...[]byte) error {
--- a/resp/resp.go
+++ b/resp/resp.go
@@ -196,10 +196,12 @@ func bufferedIntDelim(br *bufio.Reader) (int64, error) {
 
 func readAllAppend(r io.Reader, b []byte) ([]byte, error) {
 	buf := getBuffer()
-	// a side effect of this is that the given b will be re-allocated if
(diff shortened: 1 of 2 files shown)


nussjustin:
More context on this: I have a program that reads a set of hash keys and calls HGETALL for each key in a pipeline. Avoiding the allocations in readAllAppend improves performance by at least 10% even for small numbers of keys.

Benchmarks for different set sizes:

```
name            old time/op    new time/op    delta
ReadAll/1-8        123µs ± 3%     117µs ± 5%     ~     (p=0.056 n=5+5)
ReadAll/10-8       507µs ± 3%     452µs ± 4%  -10.88%  (p=0.008 n=5+5)
ReadAll/100-8     4.88ms ± 3%    3.76ms ± 1%  -22.92%  (p=0.008 n=5+5)
ReadAll/1000-8    50.7ms ± 1%    44.7ms ± 2%  -11.94%  (p=0.008 n=5+5)

name            old alloc/op   new alloc/op   delta
ReadAll/1-8       11.5kB ± 0%     1.2kB ± 0%  -89.43%  (p=0.000 n=5+4)
ReadAll/10-8       122kB ± 0%      19kB ± 0%  -84.21%  (p=0.008 n=5+5)
ReadAll/100-8     1.22MB ± 0%    0.18MB ± 0%  -84.95%  (p=0.008 n=5+5)
ReadAll/1000-8    12.3MB ± 0%     1.9MB ± 0%  -84.21%  (p=0.008 n=5+5)

name            old allocs/op  new allocs/op  delta
ReadAll/1-8          161 ± 0%        69 ± 0%  -57.14%  (p=0.008 n=5+5)
ReadAll/10-8       1.62k ± 0%     0.70k ± 0%  -56.86%  (p=0.008 n=5+5)
ReadAll/100-8      16.2k ± 0%      7.0k ± 0%  -56.78%  (p=0.008 n=5+5)
ReadAll/1000-8      162k ± 0%       70k ± 0%  -56.64%  (p=0.008 n=5+5)
``` 

The benchmark only include the HGETALL commands not getting the keys.

resp/resp.go
```diff
@@ -179,12 +195,17 @@ func bufferedIntDelim(br *bufio.Reader) (int64, error) {
 }
 
 func readAllAppend(r io.Reader, b []byte) ([]byte, error) {
-	buf := bytes.NewBuffer(b)
+	buf := getBuffer()
 	// a side effect of this is that the given b will be re-allocated if
```
mediocregopher: I believe this comment is no longer correct, right? B might still get reallocated if its cap is too low, but it doesn't have anything to do with MinRead anymore

nussjustin: You're right. I updated the comment in the latest commit.

PR #22, opened 2018-08-02
Use JoinHostPort for IPs from sentinel

nussjustin:
IPv6 addresses need special handling when joining with a port. Although probably a rare case\*, we can just use net.JoinHostPort which accounts for this.

\* In fact I just started a sentinel server just to check if it even supports IPv6 and yes, it does.

### DIFF
--- a/sentinel.go
+++ b/sentinel.go
@@ -2,10 +2,10 @@ package radix
 
 import (
 	"errors"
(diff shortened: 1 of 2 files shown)


mediocregopher:
Good catch, thanks!

PR #23, opened 2018-08-13
Fix AUTH example

mgood:
This was passing "AUTH" as the first parameter where the receiver should be, so this was failing with an error that "mySuperSecretPassword" is an invalid command.


mediocregopher:
Good catch, thanks!

PR #25, opened 2018-08-16
Add support for streams in Redis Cluster

nussjustin:
Redis 5 (currently at RC4) adds support for [streams](https://redis.io/topics/streams-intro), including the new commands [XREAD](https://redis.io/commands/xread) and [XREADGROUP](https://redis.io/commands/xreadgroup).

Those two commands are special in that the stream keys are defined after a `STREAMS` argument and are followed by an equal number of stream IDs. The other streams command all take just a single key and as the first argument, like most other Redis commands.

The [Keys](https://github.com/mediocregopher/radix.v3/blob/461cce078b93bb7f31b791af913c1e6e07a86053/action.go) method of `cmdAction` currently assumes that the key is always the first argument, except for the [BITOP](https://redis.io/commands/bitop) command which has special handling. For `XREAD` and `XREADGROUP` to work they would need special handling like `BITOP` (although a tiny bit more complex).

This commit adds support for `XREAD` and `XREADGROUPS`, including optional testing of the commands for Redis 5 (the tests check if streams are supported and won't send the commands to Redis when they are not supported).

Note that without Redis Cluster (e.g. a standalone Redis or Sentinel) the commands will already work correctly and this PR only affects Redis Cluster use and external uses of the `Keys` method.

### DIFF
--- a/action.go
+++ b/action.go
@@ -153,10 +153,29 @@ func Cmd(rcv interface{}, cmd string, args ...string) CmdAction {
 	}
 }
 
+func findStreamsKeys(args []string) []string {
+	for i, arg := range args {
+		if strings.ToUpper(arg) != "STREAMS" {
+			continue
+		}
+
+		// after STREAMS only stream keys and IDs can be given and since there must be the same number of keys and ids
+		// we can just take half of remaining arguments as keys. If the number of IDs does not match the number of
+		// keys the command will fail later when send to Redis so no need for us to handle that case.
+		ids := len(args[i+1:]) / 2
+
+		return args[i+1:len(args)-ids]
+	}
+
+	return nil
+}
+
 func (c *cmdAction) Keys() []string {
 	cmd := strings.ToUpper(c.cmd)
 	if cmd == "BITOP" && len(c.args) > 1 { // antirez why you do this
 		return c.args[1:]
+	} else if cmd == "XREAD" || cmd == "XREADGROUP" { // antirez why you still do this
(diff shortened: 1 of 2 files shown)


nussjustin:
@mediocregopher This is basically a working proof of concept for streams support that I'd like to get feedback on.

The changes to the `cmdAction` type are quite minimal and most of the code is tests.

Since Redis 5 is not released yet and Travis runs Redis 4, the commands themselfs won't be executed by Travis yet, but I verified everything locally.

mediocregopher:
Dang, I had forgotten streams were gonna be a thing, that's exciting they're gonna be out soon! Just read through the stream doc and I gotta say that it's not the prettiest API that antirez ever came up with XD but your code looks alright to me. Only note I would make is to add a TODO comment by the tests to get rid of that check once 5 goes stable and travis starts using it. Other than that great PR as always :P

nussjustin:
Done. 

Yeah stream support will be awesome :-) Although I agree that the API is not the prettiest (but still pretty compared to the changes in the [RESP3](https://gist.github.com/antirez/2bc68a9e9e45395e297d288453d5d54c) draft, IMO)

mediocregopher:
I hadn't even seen RESP3 yet, dang, I should get this package promoted to be the recommended soon before I might have to rewrite it again for resp3 support >_<

nussjustin:
@mediocregopher RESP3 is still draft but it will need **a lot** of changes. But yeah, RESP3 may call for a radix.v4 :-/

PR #26, opened 2018-08-18
Preallocate space in new maps in resp.Any

nussjustin:
Possible since 1.9, avoids a few allocations in most cases


mediocregopher:
👌

PR #35, opened 2018-10-07
Replace readAllAppend and limitedReaderPlus with more efficient readNAppend

nussjustin:
When ever `readAllAppend` is called it receives an `*limitedReaderPlus`, which always ends after an specific amount of bytes that we have specified before. This change makes use of this fact and replaces `readAllAppend` and the `limitedReaderPlus` type with a new function `readNAppend` that reads a given number of bytes. This avoids some allocation and memory movement and removes the need for copying into a `*bytes.Buffer` first and also removes an usage of `sync.Pool`.

Benchmark:

```
name                       old time/op    new time/op    delta
BenchmarkReadNAppend/N=0-8       65.8ns ± 0%    18.5ns ± 0%  -71.90%  (p=0.000 n=10+8)
BenchmarkReadNAppend/N=64-8      80.2ns ± 3%    39.8ns ± 0%  -50.34%  (p=0.000 n=10+9)
BenchmarkReadNAppend/N=512-8      253ns ± 2%     206ns ± 2%  -18.64%  (p=0.000 n=9+10)
BenchmarkReadNAppend/N=4096-8    1.09µs ± 3%    1.04µs ± 5%   -5.22%  (p=0.001 n=10+10)

name                       old alloc/op   new alloc/op   delta
BenchmarkReadNAppend/N=0-8        0.00B          0.00B          ~     (all equal)
BenchmarkReadNAppend/N=64-8       0.00B          0.00B          ~     (all equal)
BenchmarkReadNAppend/N=512-8       512B ± 0%      512B ± 0%     ~     (all equal)
BenchmarkReadNAppend/N=4096-8    4.10kB ± 0%    4.10kB ± 0%   -0.02%  (p=0.000 n=10+10)

name                       old allocs/op  new allocs/op  delta
BenchmarkReadNAppend/N=0-8         0.00           0.00          ~     (all equal)
BenchmarkReadNAppend/N=64-8        0.00           0.00          ~     (all equal)
BenchmarkReadNAppend/N=512-8       1.00 ± 0%      1.00 ± 0%     ~     (all equal)
BenchmarkReadNAppend/N=4096-8      1.00 ± 0%      1.00 ± 0%     ~     (all equal)
```


nussjustin:
The radix GetSet benchmarks show no changes:

```
name                     old time/op    new time/op    delta
SerialGetSet/radix-8       43.9µs ±10%    44.5µs ± 2%    ~     (p=0.897 n=10+8)
ParallelGetSet/radix-8     13.2µs ± 4%    13.4µs ± 6%    ~     (p=0.247 n=10+10)

name                     old alloc/op   new alloc/op   delta
SerialGetSet/radix-8        67.0B ± 0%     67.0B ± 0%    ~     (all equal)
ParallelGetSet/radix-8       100B ± 0%       99B ± 0%  -1.00%  (p=0.000 n=10+9)

name                     old allocs/op  new allocs/op  delta
SerialGetSet/radix-8         4.00 ± 0%      4.00 ± 0%    ~     (all equal)
ParallelGetSet/radix-8       5.00 ± 0%      5.00 ± 0%    ~     (all equal)
```

nussjustin:
A local test for GET and MGET with big values shows a great improvement of >36%

```
name         old time/op    new time/op    delta
RedisGet-8      106ms ± 7%      62ms ± 5%  -41.17%  (p=0.000 n=10+10)
RedisMGet-8    87.4ms ± 8%    55.8ms ± 7%  -36.15%  (p=0.000 n=10+10)

name         old alloc/op   new alloc/op   delta
RedisGet-8      513MB ±13%     220MB ± 6%  -57.12%  (p=0.000 n=10+10)
RedisMGet-8     353MB ±20%     161MB ± 8%  -54.41%  (p=0.000 n=10+10)

name         old allocs/op  new allocs/op  delta
RedisGet-8        112 ±16%        75 ± 4%  -33.07%  (p=0.000 n=10+10)
RedisMGet-8      52.7 ±15%      32.4 ± 5%  -38.44%  (p=0.000 n=10+9)
```

The code can be found here: https://gist.github.com/nussjustin/7485812ce6303c8980b88b07f26b3474

mediocregopher:
This looks pretty awesome at first glance, lemme take a better look at it tomorrow and make sure I understand it properly before merging

nussjustin:
Yeah, while the idea is quiet simple the implementation is quite tricky and took me some time to get right. There's probably still area to improvement, especially with the io.Reader handling (0 byte reading).

But given the performance improvement for my gist, the implementation looked like a great tradeoff when I tested it.

resp/resp.go
```diff
@@ -204,12 +214,26 @@ func readAllAppend(r io.Reader, b []byte) ([]byte, error) {
...
+	// ReadFull does not call Read if n == 0, but some readers (like
+	// limitedReaderPlus) need at least one call to Read, so we just
+	// call Read with an empty []byte if n == 0.
+	if n == 0 {
+		_, err := r.Read(b[len(b):])
+		return b, err
+	} else if m := cap(b) - len(b); n > m {
+		// Go 1.11 optimizes append(x, make([]T, y)...) calls by
+		// avoiding unnecessary allocations and copies. (CL 109517)
+		b = append(b[:cap(b)], make([]byte, n-m)...)[:len(b)]
```
mediocregopher: In this file there's already a function called `expand`, which is used other places. I _think_ this logic could be moved into there, and that function could then be used here too, like:

```
b = expand(b, len(b)+n)
_, err := io.ReadFull(r, b[len(b):len(b)+n])
return b, err
```

nussjustin:
Please take a look. Commit 3 to 5 address your comments. The last 2 commits (ok, the second last) removes `readAllAppend` and `limitedReaderPlus` as you suggested.

nussjustin:
The newest code gives even more improvement on my gist:

```
name         old time/op    new time/op    delta
RedisGet-8      124ms ± 2%      64ms ± 7%  -48.61%  (p=0.000 n=8+10)
RedisMGet-8     101ms ±14%      59ms ± 3%  -41.37%  (p=0.000 n=10+10)

name         old alloc/op   new alloc/op   delta
RedisGet-8      512MB ± 9%     210MB ± 7%  -58.97%  (p=0.000 n=10+10)
RedisMGet-8     350MB ±24%     163MB ± 8%  -53.47%  (p=0.000 n=10+9)

name         old allocs/op  new allocs/op  delta
RedisGet-8        110 ±11%        70 ± 4%  -36.13%  (p=0.000 n=10+9)
RedisMGet-8      52.2 ±23%      31.7 ± 4%  -39.34%  (p=0.000 n=10+9)
```

mediocregopher:
Nice! Thanks so much for all that, these speed improvements are amazing!

mediocregopher:
Ack! actually I did have one question, forgot to ask, on line 804:

```
if err := a.unmarshalSingle(br, int(l)); err != nil && err != io.EOF {
```

What's the purpose of allowing `io.EOF`? if that was hit before the the trailing `\r\n` were discarded I would think that error should be passed up

nussjustin:
Well... Yeah... That makes more sense. #38 :-)

PR #38, opened 2018-10-11
Return unexpected io.EOF for bulk strings to the user

nussjustin:
If we encounter an io.EOF before the bulk string ends we should report it to the user instead of assuming that the Discard call returns io.EOF too.


mediocregopher:
Lol, perfect

PR #39, opened 2018-10-11
Use net.Buffers for multiWrite

nussjustin:
Simplifies multiWrite a bit and helps in cases when a user writes directly to a net.Conn, by avoiding some syscalls.

### DIFF
--- a/resp/resp.go
+++ b/resp/resp.go
@@ -13,6 +13,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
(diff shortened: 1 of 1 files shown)


nussjustin:
I'm using the resp package directly in one of my small import scripts / tools and this helps a bit since I'm writing directly to a net.Conn. 

No performance numbers because I don't have my other notebook at hand.

mediocregopher:
Neat! Didn't know that was a thing

PR #41, opened 2018-10-19
Add support for XGROUP and XINFO in Redis Cluster

nussjustin:
With Redis 5.0 released (finally!) the remaining stream commands are also now finalized and we need to add support for XINFO and XGROUP. XINFO is again a bit special, since one case (XINFO HELP) does not take any key.


mediocregopher:
👍 thanks!

Issue #56, opened 2018-12-14
Sentinel Client method leaks secondary pools


nussjustin:
Disclaimer: This is all theoretical and I haven't tested it. I only looked at the code and stepped through this scenario in my head.

The [Sentinel.Client](https://godoc.org/github.com/mediocregopher/radix#Sentinel.Client) method says

> NOTE the Client should _not_ be closed.

This can be a problem since Clients to secondaries are created by calling the [SentinelPoolFunc](https://godoc.org/github.com/mediocregopher/radix#SentinelPoolFunc), which is primarily used for creating a long lived pool for connections to the primary and has the same configuration for primaries and secondaries.

If the returned pool for one of the secondaries is not closed, the pool can leak connections (at least for some time, depending on the pool configuration/implementation). But users don't know this and are told to not close the returned Client.

Note that for the case where an user wants a client for the primary they really must not call Close.

This doesn't seem like a big problem to me, since I don't expect many people to manually use connections to secondaries, but is still worth fixing.

I thought about two ways to solve this, but there are probably others:

1. Save (and reuse) the non-primary pools and close them when necessary (on Master change).
2. Always return a fresh Client, using the [SentinelConnFunc](https://godoc.org/github.com/mediocregopher/radix#SentinelConnFunc), even for the primary and tell users to manually close the connections.

nussjustin:
My concrete idea for 1 would be to changes `sc.secsM` to a `map[string]Client` and then add `nil` Clients for new secondaries in `setClients` and `.Close` + `delete` old secondaries from `sc.secsM`. This could even be extended so that all connections (master + secondaries) are handled the same way and `Do` would just basically just do `sc.Client(sc.primAddr)` to get the master connection.

mediocregopher:
Hey @nussjustin, thanks for pointing this out! You're definitely right that the current behavior isn't ideal, I guess I got lazy while implementing what's there. I like your idea to keep open Clients in `secsM`, judging by the TODO I left inside the `Client` method that was my original intention, I just didn't do it >_<

I'll try to tackle this later tonight.

PR #57, opened 2018-12-21
Sentinel fixes

mediocregopher:
@nussjustin here's a potential fix for #56, still needs tests but I'll do that later this week

### DIFF
--- a/sentinel.go
+++ b/sentinel.go
@@ -287,13 +287,12 @@ func sentinelMtoAddr(m map[string]string, cmd string) (string, error) {
 // given a connection to a sentinel, ensures that the Clients currently being
 // held agrees with what the sentinel thinks they should be
 func (sc *Sentinel) ensureClients(conn Conn) error {
-	// TODO it'd be ideal to pipeline these two commands, but doing so breaks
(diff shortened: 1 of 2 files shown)


sentinel.go
```diff
@@ -289,50 +301,93 @@ func (sc *Sentinel) ensureClients(conn Conn) error {
...
+	}
 
-	} else if !primChanged {
-		sc.l.Lock()
-		sc.secsM = newSecsM
-		sc.l.Unlock()
+	// if the primary has then record that the replica set state has changed,
+	// and while we're here try to pre-emtively populate newPrim from the old
+	// secondary set.
+	if oldPrimAddr != newPrimAddr {
```
nussjustin: There is quite a bit of code just because the primary is kept separate from all the other clients. Instead we could keep them all in one map and then just lookup the primary by `sc.prim` when needed. This way we could treat all clients the same (we would need to skip the primary in `Addrs`, but that's just a simple `if addr == sc.primAddr`)

What do you think?

mediocregopher: went ahead and did that, you're right it did simplify the code quite a bit

PR #61, opened 2019-01-01
Automatic pipelining for commands in Pool

### LINKED ISSUE #58: Check out RedisPipe and figure out if its strategy of implicit pipelining/batching can be incorporated
https://github.com/joomcode/redispipe

That project essentially uses a batching concept to implicitly pipeline all commands, and uses a WritePause timeout to batch commands together into single system reads/writes. I think doing something similar in radix _might_ be possible, though I haven't really thought out how yet. I think the main challenge will be in coordinating commands across Pool's connections in a way which utilizes the batching to its fullest extent.

nussjustin:
This adds automatic pipelining for `CmdAction`s in a Pool similar to the way redispipe works, but using the fact that a Pool has multiple connections to allow concurrent execution of commands that can't be pipelined.

The pooling is enabled by default and can be configured or even disabled by users. The default window for pipelining is 150us which is the window suggested by redispipe for production usage.

The benchmark `BenchmarkParallelGetSet/radix` was updated to test the new (`default`) behaviour and both the old behaviour (no pipelining) and limited pipelining (1 concurrent execution).

Fixes #58


pipeliner.go
```diff
@@ -0,0 +1,246 @@
...
+		resCh: make(chan error, 1),
+	}
+}
+
+func poolPipedCmd(req *pipedCmd) {
+	req.cmd = nil
+	pipedCmdPool.Put(req)
+}
+
+type pipedCmdPipeline []pipedCmd
```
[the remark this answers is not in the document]
nussjustin: Done, although it was a bit more complex than this. I had to embed `pipeline` into the `pipelinerPipeline` and change the code to work with pointers to `pipelinerCmd` to avoid allocations (since adding the `pipelinerCmd` struct to a `pipeline` involves boxing)

pool.go
```diff
@@ -171,6 +192,38 @@ func PoolOnFullBuffer(size int, drainInterval time.Duration) PoolOpt {
...
+		po.pipelineConcurrency = limit
+	}
+}
+
+// PoolPipelineLimit sets the maximum number of concurrent commands that can be
+// automatically pipelined onto a single Redis connection.
+//
+// If limit is zero then no limit will be used and pipelines will only be limited
+// by the specified time window.
+func PoolPipelineLimit(limit int) PoolOpt {
```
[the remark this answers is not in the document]
nussjustin: Done

pool.go
```diff
@@ -433,8 +506,21 @@ func (p *Pool) put(ioc *ioErrConn) {
...
-// to the pool
+// to the pool.
+//
+// If the given Action is a CmdAction, it will be pipelined with other concurrent
+// calls to Do, which can improve the performance and resource usage of the Redis
+// server, but will increase the latency for some of the Actions. To avoid the
+// automatic pipelining you can either set PoolPipelineWindow(0) when creating the
+// Pool or use WithConn. Pipelines created manually (via Pipeline) are also excluded
+// from this and will be executed as if using WithConn.
+// Due to a limitation in the implementation, custom CmdAction implementations
```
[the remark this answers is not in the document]
nussjustin: Done

pipeliner.go
```diff
@@ -0,0 +1,246 @@
...
+
+func (p *pipeliner) flush(pipe []pipedCmd) {
+	if len(pipe) == 0 {
+		return
+	}
+
+	// copy requests into a pipeline so that we can flush the pipeline in
+	// the background, to avoid blocking the main loop.
+	pipeCopy := make(pipedCmdPipeline, len(pipe))
+	copy(pipeCopy, pipe)
```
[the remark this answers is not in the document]
nussjustin: I solved this by changing the `flushSema` to contain buffers / slices for commands. The new code takes it`s `[]CmdAction` slices from the now called `reqsBufCh` and avoids both copying and allocating this way.

bench_test.go
```diff
@@ -54,41 +108,126 @@ func BenchmarkParallelGetSet(b *B) {
...
+					b.Fatal(err)
+				} else if out != "bar" {
+					b.Fatal("got wrong value")
 				}
-				return nil
-			}))
+			})
+		})
+
+		b.Run("multiple pipelines", func(b *B) {
```
mediocregopher: For the sake of looking good, let's call this `"default"` :P

pool.go
```diff
@@ -171,6 +192,31 @@ func PoolOnFullBuffer(size int, drainInterval time.Duration) PoolOpt {
...
+//
+// If limit is greater than the pool size or less than 1, the limit will be
+// set to the pool size.
+func PoolPipelineConcurrency(limit int) PoolOpt {
+	return func(po *poolOpts) {
+		po.pipelineConcurrency = limit
+	}
+}
+
+// PoolPipelineLimit sets the duration after which internal pipelines will be
```
mediocregopher: `PoolPipelineWindow`*

pool.go
```diff
@@ -15,6 +15,24 @@ var ErrPoolEmpty = errors.New("connection pool is empty")
...
+	"BRPOPLPUSH": true,
+
+	"BZPOPMIN": true,
+	"BZPOPMAX": true,
+
+	"XREAD":      true,
+	"XREADGROUP": true,
+
+	"SAVE":  true,
+	"WATCH": true,
```
funny-falcon-at-joomcode: It is worthless to watch for WATCH itself. WATCH is usable only in conjunction with following MULTI+EXEC.

pool.go
nussjustin: Done

bench_test.go
nussjustin: Done

pool.go
nussjustin: Done

funny-falcon:
You didn't fully understand my concern about WATCH command:
- WATCH could not be sent separately from following MULTI+EXEC
- that means:
  a) it could not be pipelined, because otherwise commands from other goroutines could interleave between WATCH and MULTI
  b) it could not be sent in separate connection from MULTI,

That means, WATCH could be allowed only within WithConn, and completely forbidden in any other case.

mediocregopher:
My attitude has been to only do command-specific logic when it's necessary to make the code function (eg knowing which commands have keys when directing them at the right cluster node, or knowing which block so they aren't  pipelined). Ensuring that the user doesn't do something the wrong way is a much bigger can of worms that I'm not sure is worth opening. The top level docs on the package specify that WithConn must be used for MULTI/EXEC stuff, I think that's enough for now. So with that said, WATCH shouldn't be included in the list of commands which block, since it doesn't actually cause the connection to block.

nussjustin:
I'm okay with this. The current implementation already does not try to handle WATCH in any special way.

pool.go
```diff
@@ -433,8 +481,22 @@ func (p *Pool) put(ioc *ioErrConn) {
 
 // Do implements the Do method of the Client interface by retrieving a Conn out
 // of the pool, calling Run on the given Action with it, and returning the Conn
-// to the pool
+// to the pool.
+//
+// If the given Action is a CmdAction, it will be pipelined with other concurrent
+// calls to Do, which can improve the performance and resource usage of the Redis
+// server, but will increase the latency for some of the Actions. To avoid the
+// automatic pipelining you can either set PoolPipelineWindow(0) when creating the
```
mediocregopher: `PoolPipelineWindow(0, 0)`* I swear this is the last thing and then this can go in >_<

pipeliner.go
```diff
@@ -0,0 +1,241 @@
...
+		}
+	}
+
+	return p
+}
+
+// CanDo checks if the given Action can be executed / passed to p.Do.
+//
+// If CanDo returns false, the Action must not be given to Do.
+func (p *pipeliner) CanDo(a Action) bool {
```
mediocregopher: This method could be significantly simplified:

```go
// there is currently no way to get the command for CmdAction implementations
// from outside the radix package so we can not multiplex those commands. User
// defined peipelines are not pipelined to let the user better control them.
if cmdA, ok := a.(*cmdAction); ok {
     return !blockingCmds[strings.ToUpper(cmdA.cmd)]
}
return false

nussjustin: Done

pool.go
nussjustin: Done

PR #68, opened 2019-02-14
clean-up go.mod benchmark dependencies

mwf:
1. Move bench_test.go to sub-module;
2. Wait for Pool initialization explicitly - it seems you can't use it with asynchronous init


pool.go
```diff
@@ -314,6 +331,15 @@ func NewPool(network, addr string, size int, opts ...PoolOpt) (*Pool, error) {
 	if p.opts.overflowSize > 0 && p.opts.overflowDrainInterval > 0 {
 		p.atIntervalDo(p.opts.overflowDrainInterval, p.doOverflowDrain)
 	}
+
+	// wait all connections to establish - as tests show, you can't work with pool
+	// until p.initDone is closed!
```
nussjustin: This is wrong. The Pool can be used before all connections are created. The benchmark just waits for the pool to be filled, to avoid measuring the overhead of creating connections.

Waiting for the pool to be full before returning here will also needlessly slow down many uses of Pools.

Also this is leaking the connections and goroutines.

pool.go
```diff
@@ -86,6 +93,14 @@ func PoolConnFunc(cf ConnFunc) PoolOpt {
 	}
 }
 
+// PoolDialTimeout specifies the timeout for pool initialization.
+// NewPool will fail if connections are not created after the deadline.
+func PoolDialTimeout(d time.Duration) PoolOpt {
```
nussjustin: The function name will probably confuse people since this is not about dialing redis but filling the pool.

pool.go
mwf: Thanks! Actually I've experienced some problems with tests when I tried to remove the wait on initDone.

I'll re-check it soon and come back with details.

mediocregopher:
Hi @mwf, I'm a bit confused about what this fixes, specifically. Are you having a problem with the current `go.mod` not actually working? You mentioned in the issue that the version the errorx package is pinned to is possibly broken, is that what you are trying to fix?

pool.go
mediocregopher: For clarity, _some_ tests do rely on the Pool to be full before they run, hence the need for initDone in the first place. But in general we don't want the Pool to be waiting for all connections to be created before returning.

mwf:
Hi @mediocregopher!

This fixes the dependency graph in first place.
Please try to do the following:

```
cd `mktemp -d`
go mod init example.com/test
cat <<EOD > main.go 
package main

import (
        _ "github.com/mediocregopher/radix"
)

func main() {}

EOD

go mod tidy
go build .
```

Build passes OK, no questions here.
But the go.mod and the dependency graph contains the projects I don't expect to see there.

go.mod
```
module example.com/test

require (
	github.com/garyburd/redigo v1.6.0 // indirect
	github.com/mediocregopher/radix.v2 v0.0.0-20181115013041-b67df6e626f9 // indirect
	github.com/mediocregopher/radix/v3 v3.2.0
)
```
The indirects are there because you haven't done `go mod tidy` in your project.

And the full graph will be:
```
$ go mod graph
example.com/test github.com/garyburd/redigo@v1.6.0
example.com/test github.com/mediocregopher/radix.v2@v0.0.0-20181115013041-b67df6e626f9
example.com/test github.com/mediocregopher/radix/v3@v3.2.0
github.com/mediocregopher/radix/v3@v3.2.0 github.com/davecgh/go-spew@v1.1.1
github.com/mediocregopher/radix/v3@v3.2.0 github.com/gomodule/redigo@v2.0.0+incompatible
github.com/mediocregopher/radix/v3@v3.2.0 github.com/joomcode/errorx@v0.1.0
github.com/mediocregopher/radix/v3@v3.2.0 github.com/joomcode/redispipe@v0.9.0
github.com/mediocregopher/radix/v3@v3.2.0 github.com/mediocregopher/mediocre-go-lib@v0.0.0-20181029021733-cb65787f37ed
github.com/mediocregopher/radix/v3@v3.2.0 github.com/pmezard/go-difflib@v1.0.0
github.com/mediocregopher/radix/v3@v3.2.0 github.com/stretchr/testify@v1.2.2
```

I don't like the _idea_ that in order to download `radix/v3` I have to download some unrelated stuff just because it is used in project benchmarks.

Especially funny that `example.com/test github.com/mediocregopher/radix.v2` is downloaded, because it is used in `github.com/joomcode/redispipe` benchmarks :))
Looks like a dependency hell :)

So I just tried to extract your benchmarks to a separate go module, in order to clean-up the **real** library dependencies.

But the fact that benchmarks need a private channel to start correctly ruins this idea a little bit.
So there are 2 solutions:
1. Extract only 3rd party benchmarks to a separate module, keeping your benchmarks as is. But then it becomes hard to run them.
2. Get rid of private field access in some way. 

So I'd be glad to implement any of this solutions, but a decision should be made.

I completely agree that blocking on filling the pool is a bad idea, actually it was just a workaround to get the tests passed. And a kind of a question to you, because I understood that must not be the intended behaviour :)

Hope I made my intentions clear, thanks for your time!

nussjustin:
> But the fact that benchmarks need a private channel to start correctly ruins this idea a little bit.

It's not great but without this we are just measuring a cold pool which would not represent most actual users, since mostly pools will be long lived.
 
The only other quick way around this, that currently comes to my mind, is executing a bunch of commands before the benchmarks start to force the pools to be filled. This would work across all Redis clients, but is not really that clean.

mwf:
> The only other quick way around this, that currently comes to my mind, is executing a bunch of commands before the benchmarks start to force the pools to be filled. This would work across all Redis clients, but is not really that clean.

Another possible solution is to implement some stats exporter like https://golang.org/pkg/database/sql/#DB.Stats

And we could either poll the connection count in pool or implement a method, returning this "initDone" channel.

Such stats could be pretty helpful for monitoring on client side.

mediocregopher:
> This fixes the dependency graph in first place.

"fixes" is a strong word, since the graph isn't technically broken; "simplifies" is maybe better :P 

I'm not opposed to simplifying the dependency graph by moving the benchmarks into a sub-module, but I don't think it's worth adding something to the API to do so.

> And we could either poll the connection count in pool or implement a method, returning this "initDone" channel.

Pool already has `NumAvailConns`, so polling that would work.

mwf:
> "fixes" is a strong word, since the graph isn't technically broken; "simplifies" is maybe better :P

Oh, yeap 👍 

Thank you for the comments, guys! I've done the `NumAvailConns` polling - it fits perfectly.

If you'd like I could squash my commits into a single one before merging, if you're OK with the suggested changes.

mediocregopher:
I went ahead and squashed it via github, I'll tag this `v3.2.0` since I apparently forgot to tag all the previous PRs.

Thanks for the PR and all the follow-through! Sorry you had to undo most of your work >_< For the future, waiting for a response to the issue before starting the PR can help prevent throwing away a bunch of work.

mwf:
> Sorry you had to undo most of your work
No problem, I assumed I was doing something wrong, but I needed some proof-of-concept anyway :)

Sometimes making a wrong PR with a good review is faster than waiting for a response in the issue.

Thanks for merging :)

PR #71, opened 2019-02-19
Optimize command encoding and decoding a bit 

nussjustin:
While benchmarking today I found that we spend quite a lot of time in `defer`s and calls to `bytesutil.MultiWrite` as well as discarding data.

Most of the `defer`s and `bytes.MultiWrite` calls come from encoding the commands, which currently need `2+len(args)` `defer`s and `bytes.MultiWrite` calls per command.

The overhead in discarding data comes from passing a `*bytes.Reader` to `bytesutil.ReadNDiscard` which can currently not use a fast-path.

This PR fixes both cases and improves performance quite a bit for the pipelined benchmarks. The non-pipelined benchmarks don't improve, and are not as stable as the pipelined benchmarks, since they are mostly constrained by syscall overhead in both radix and Redis.

```
name                                   old time/op    new time/op    delta
SerialGetSet/radix-16                    45.0µs ± 2%    43.7µs ± 1%   -2.94%  (p=0.008 n=5+5)
ParallelGetSet/radix/no_pipelining-16    16.0µs ± 7%    17.5µs ± 5%   +9.44%  (p=0.016 n=5+5)
ParallelGetSet/radix/one_pipeline-16     6.16µs ± 0%    4.85µs ± 0%  -21.26%  (p=0.008 n=5+5)
ParallelGetSet/radix/default-16          4.14µs ± 1%    3.43µs ± 0%  -17.22%  (p=0.008 n=5+5)

name                                   old alloc/op   new alloc/op   delta
SerialGetSet/radix-16                     67.0B ± 0%     67.0B ± 0%     ~     (all equal)
ParallelGetSet/radix/no_pipelining-16      116B ± 0%      116B ± 0%     ~     (all equal)
ParallelGetSet/radix/one_pipeline-16      70.0B ± 0%     70.0B ± 0%     ~     (all equal)
ParallelGetSet/radix/default-16           71.0B ± 0%     72.0B ± 0%   +1.41%  (p=0.008 n=5+5)

name                                   old allocs/op  new allocs/op  delta
SerialGetSet/radix-16                      4.00 ± 0%      4.00 ± 0%     ~     (all equal)
ParallelGetSet/radix/no_pipelining-16      6.00 ± 0%      6.00 ± 0%     ~     (all equal)
ParallelGetSet/radix/one_pipeline-16       4.00 ± 0%      4.00 ± 0%     ~     (all equal)
ParallelGetSet/radix/default-16            4.00 ± 0%      4.00 ± 0%     ~     (all equal)
```

I haven't looked at the `ParallelGetSet/radix/no_pipelining` regression, but it was probably a problem with my test system.

### DIFF
--- a/bench/bench_test.go
+++ b/bench/bench_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"runtime"
+	"strings"
 	. "testing"
 	"time"
 
@@ -90,6 +91,84 @@ func BenchmarkSerialGetSet(b *B) {
 				b.Fatal(res)
 			}
 			if res := sync.Do("GET", "foo"); redis.AsError(res) != nil {
+				b.Fatal(res)
+			}
+		}
+	})
+}
+
+func BenchmarkSerialGetSetLargeArgs(b *B) {
+	key := strings.Repeat("foo", 24)
+	val := strings.Repeat("bar", 4096)
+
+	b.Run("radix", func(b *B) {
+		rad, err := radix.Dial("tcp", "127.0.0.1:6379")
+		if err != nil {
+			b.Fatal(err)
+		}
+		defer rad.Close()
+		b.ResetTimer()
+		for i := 0; i < b.N; i++ {
+			if err := rad.Do(radix.Cmd(nil, "SET", key, val)); err != nil {
+				b.Fatal(err)
+			}
+			var out string
+			if err := rad.Do(radix.Cmd(&out, "GET", key)); err != nil {
+				b.Fatal(err)
+			}
+		}
+	})
+
+	b.Run("redigo", func(b *B) {
+		red := newRedigo()
+		b.ResetTimer()
+		for i := 0; i < b.N; i++ {
+			if _, err := red.Do("SET", key, val); err != nil {
+				b.Fatal(err)
+			}
+			if _, err := redigo.String(red.Do("GET", key)); err != nil {
+				b.Fatal(err)
+			}
+		}
+	})
+
+	b.Run("redispipe", func(b *B) {
+		pipe, err := redisconn.Connect(context.Background(), "127.0.0.1:6379", redisconn.Opts{
+			Logger:     redisconn.NoopLogger{},
+			WritePause: 150 * time.Microsecond,
+		})
+		defer pipe.Close()
+		if err != nil {
+			b.Fatal(err)
+		}
+		sync := redis.Sync{pipe}
+		b.ResetTimer()
+		for i := 0; i < b.N; i++ {
+			if res := sync.Do("SET", key, val); redis.AsError(res) != nil {
+				b.Fatal(res)
+			}
+			if res := sync.Do("GET", key); redis.AsError(res) != nil {
+				b.Fatal(res)
+			}
+		}
+	})
+
+	b.Run("redispipe_pause0", func(b *B) {
+		pipe, err := redisconn.Connect(context.Background(), "127.0.0.1:6379", redisconn.Opts{
+			Logger:     redisconn.NoopLogger{},
+			WritePause: -1,
+		})
+		defer pipe.Close()
+		if err != nil {
+			b.Fatal(err)
+		}
+		sync := redis.Sync{pipe}
+		b.ResetTimer()
+		for i := 0; i < b.N; i++ {
+			if res := sync.Do("SET", key, val); redis.AsError(res) != nil {
+				b.Fatal(res)
+			}
+			if res := sync.Do("GET", key); redis.AsError(res) != nil {
 				b.Fatal(res)
 			}
 		}
--- a/internal/bytesutil/bytesutil.go
+++ b/internal/bytesutil/bytesutil.go
@@ -175,8 +175,14 @@ func ReadNDiscard(r io.Reader, n int) error {
 
 	if n == 0 {
 		return nil
-	} else if d, ok := r.(discarder); ok {
-		_, err := d.Discard(n)
+	}
+
+	switch v := r.(type) {
+	case discarder:
+		_, err := v.Discard(n)
+		return err
+	case io.Seeker:
+		_, err := v.Seek(int64(n), io.SeekCurrent)
 		return err
 	}
 
--- a/resp/resp2/resp.go
+++ b/resp/resp2/resp.go
@@ -143,12 +143,18 @@ type BulkStringBytes struct {
 // MarshalRESP implements the Marshaler method
 func (b BulkStringBytes) MarshalRESP(w io.Writer) error {
 	if b.B == nil && !b.MarshalNotNil {
-		return bytesutil.MultiWrite(w, nilBulkString)
+		_, err := w.Write(nilBulkString)
+		return err
 	}
 	scratch := bytesutil.GetBytes()
-	defer bytesutil.PutBytes(scratch)
+	*scratch = append(*scratch, bulkStrPrefix...)
 	*scratch = strconv.AppendInt(*scratch, int64(len(b.B)), 10)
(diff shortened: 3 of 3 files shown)


nussjustin:
Benchmark with `-cpu 8` show smaller improvements but no drastic change for the `no_pipelining` benchmark (which reinforces my guess that it was a hiccup on my test system):

```
name                                  old time/op    new time/op    delta
SerialGetSet/radix-8                    43.9µs ± 1%    44.2µs ± 1%     ~     (p=0.222 n=5+5)
ParallelGetSet/radix/no_pipelining-8    15.4µs ± 6%    15.6µs ± 7%     ~     (p=0.548 n=5+5)
ParallelGetSet/radix/one_pipeline-8     19.0µs ± 5%    17.2µs ± 2%   -9.76%  (p=0.008 n=5+5)
ParallelGetSet/radix/default-8          16.1µs ± 4%    14.1µs ± 4%  -12.17%  (p=0.008 n=5+5)

name                                  old alloc/op   new alloc/op   delta
SerialGetSet/radix-8                     67.0B ± 0%     67.0B ± 0%     ~     (all equal)
ParallelGetSet/radix/no_pipelining-8      115B ± 0%      115B ± 0%     ~     (all equal)
ParallelGetSet/radix/one_pipeline-8      69.4B ± 1%     69.6B ± 1%     ~     (p=1.000 n=5+5)
ParallelGetSet/radix/default-8           70.0B ± 0%     70.0B ± 0%     ~     (all equal)

name                                  old allocs/op  new allocs/op  delta
SerialGetSet/radix-8                      4.00 ± 0%      4.00 ± 0%     ~     (all equal)
ParallelGetSet/radix/no_pipelining-8      6.00 ± 0%      6.00 ± 0%     ~     (all equal)
ParallelGetSet/radix/one_pipeline-8       4.00 ± 0%      4.00 ± 0%     ~     (all equal)
ParallelGetSet/radix/default-8            4.00 ± 0%      4.00 ± 0%     ~     (all equal)
```

nussjustin:
Rerunning the `no_pipelining` benchmark still gives a small performance decrease, but much less. Still high variation in ns/op for each run.

```
name                                   old time/op    new time/op    delta
ParallelGetSet/radix/no_pipelining-16    16.0µs ± 7%    16.8µs ± 9%   ~     (p=0.310 n=5+5)

name                                   old alloc/op   new alloc/op   delta
ParallelGetSet/radix/no_pipelining-16      116B ± 0%      116B ± 0%   ~     (all equal)

name                                   old allocs/op  new allocs/op  delta
ParallelGetSet/radix/no_pipelining-16      6.00 ± 0%      6.00 ± 0%   ~     (all equal)
```

mediocregopher:
Hmm that's super interesting! I didn't know that `defer` had any performance impact at all, I thought it was more of a compile-time thing, but I guess it must be putting something on the stack or something. I also wouldn't have expected there to be any change between using MultiWrite vs appending it all to a scratch and using Write, since it's all going into a `bufio.Writer` anyway. But the benchmarks don't lie :P 

Is there any reason to use MultiWrite at all anymore? If using it is inherently worse than using a scratch I'd rather just do that everywhere. If you don't feel like removing it here I can make an issue to remind me to go back and do it when I get home.

nussjustin:
> I didn't know that `defer` had any performance impact at all, I thought it was more of a compile-time thing, but I guess it must be putting something on the stack or something.

There is a small blog post here that shows the small overhead of defer: https://medium.com/i0exception/runtime-overhead-of-using-defer-in-go-7140d5c40e32

Normally one wouldn't need to think about it, but if you have many of them (and we have), there combined cost becomes easily measurable (just run the default parallel benchmark with -cpuprofile and you will see a few `runtime.defer*` functions). 

> I also wouldn't have expected there to be any change between using MultiWrite vs appending it all to a scratch and using Write, since it's all going into a `bufio.Writer` anyway.

Avoiding interface calls and packing the arguments into a slice (for the variable arguments of `bytes.MultiWrite`) helps tiny bit, but most of it comes from the `*bufio.Writer` not having to check it's available space and decide whether to flush multiple times.

There is a cost though to appending everything to `*scratch`. For small commands like in our benchmarks this cost is less than what we gain but for bigger values (e.g. when setting big strings) appending could be worse than multiple writes and eat more memory and require more allocations (since `*scratch` is rather small).

Maybe we should add a benchmark for this too (`BenchmarkLargeStringArgs`, because naming is hard? ;-)) and see if the overhead is "ok" or if we should keep a special case for large strings or strings larger than what fits into the `*scratch` buffer.

> Is there any reason to use MultiWrite at all anymore? If using it is inherently worse than using a scratch I'd rather just do that everywhere. If you don't feel like removing it here I can make an issue to remind me to go back and do it when I get home.

I wanted to keep this change small and since the other `MarshalRESP` methods are basically never called they won't affect most users. I only updated `BulkStringBytes` as part of this change for symmetry with `BulkString`.

mediocregopher:
I had some free time so I tested each of the three optimizations individually and they all definitely help, the defer one helps a surprising amount even.

> Maybe we should add a benchmark for this too (`BenchmarkLargeStringArgs`, because naming is hard? ;-)) and see if the overhead is "ok" or if we should keep a special case for large strings or strings larger than what fits into the `*scratch` buffer.

How about `SerialGetSetLargeArgs`, so we can do it for each driver too if we want. But yeah, we might need to do some shenanigans for those which are larger than scratch, it's definitely worth testing out.

> I wanted to keep this change small and since the other `MarshalRESP` methods are basically never called they won't affect most users.

That's fine, I can do the rest at a later time if I get the gumption to do it.

nussjustin:
Turns out we don't really lose performance for big strings, since `(resp2.BulkString).MarshalRESP` was already appending the string to the `*scratch` slice since it had to convert it to a `[]byte` anyway for `bytesutil.MultiWrite`. Also since we pool our buffers, we often don't need to do any allocations (especially in a non-concurrent program, where we always reuse the same `*scratch` slice).

I added the benchmark and rebased the branch onto master. Here are the results on my notebook:

```
name                                      time/op
SerialGetSetLargeArgs/radix-8             66.8µs ± 0%
SerialGetSetLargeArgs/redigo-8            64.3µs ± 1%
SerialGetSetLargeArgs/redispipe-8          500µs ± 1%
SerialGetSetLargeArgs/redispipe_pause0-8  62.5µs ± 0%

name                                      alloc/op
SerialGetSetLargeArgs/radix-8             12.5kB ± 0%
SerialGetSetLargeArgs/redigo-8            24.7kB ± 0%
SerialGetSetLargeArgs/redispipe-8         13.8kB ± 0%
SerialGetSetLargeArgs/redispipe_pause0-8  13.8kB ± 0%

name                                      allocs/op
SerialGetSetLargeArgs/radix-8               4.00 ± 0%
SerialGetSetLargeArgs/redigo-8              8.00 ± 0%
SerialGetSetLargeArgs/redispipe-8           11.0 ± 0%
SerialGetSetLargeArgs/redispipe_pause0-8    11.0 ± 0%
```

radix isn't too slow, probably because we reuse our buffers and thus avoid allocations after the first benchmark iteration.

We could still fall back to multiple writes when `*scratch` is not big enough to avoid allocations (mostly in cases where `*scratch` is often freshly allocated or just too small), but then we would also need to define some cut off length at which we do this, since allocating a new, bigger slice can still often be faster than multiple writes and amortizes over time as long as the buffer is reused.

We could also just ignore this for now, until it becomes obvious that this is a problem. Since most uses are probably more or less consistent in the size of string arguments, the pooling should work good enough for them.

mediocregopher:
> Since most uses are probably more or less consistent in the size of string arguments, the pooling should work good enough for them.

That's true, for most apps the arg sizes should stay roughly consistent.

Thanks for the PR! :D

PR #72, opened 2019-02-21
cleanup benchmark code a bit

mediocregopher:
@nussjustin I did a little cleanup of the benchmarking code to reduce some of the code duplication. One question about it though: in the `ParallelGetSet/radix/no pipeline` benchmark we'd had it using a `WithConn` within the `do`, do you remember if/why that was necessary? I took it out here because I couldn't think why it would matter, except to slow down the `no pipeline` test a bit.

### DIFF
--- a/bench/bench_test.go
+++ b/bench/bench_test.go
@@ -2,15 +2,15 @@ package bench
 
 import (
 	"context"
-	"fmt"
+	"errors"
 	"runtime"
 	"strings"
 	. "testing"
 	"time"
 
 	redigo "github.com/gomodule/redigo/redis"
-	"github.com/joomcode/redispipe/redis"
-	"github.com/joomcode/redispipe/redisconn"
+	redispipe "github.com/joomcode/redispipe/redis"
+	redispipeconn "github.com/joomcode/redispipe/redisconn"
 	"github.com/mediocregopher/radix/v3"
 )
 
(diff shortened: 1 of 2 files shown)


nussjustin:
I think it was there from the beginning (read: before I came in and had my fun with the code) and I then kept it because it disabled automatic pipelining and because I was working on the pipeliner code. But yeah, since then it's basically unnecessary and I planned to remove it at some point (it always bugged me when I looked at the code), but I'm lazy sooo....

bench/bench_test.go
```diff
@@ -187,157 +174,98 @@ func BenchmarkParallelGetSet(b *B) {
...
+				// wait for the pool to fill up
+				for {
+					time.Sleep(50 * time.Millisecond)
+					if pool.NumAvailConns() >= poolSize {
+						break
+					}
+				}
+
+				do(b, func() error {
+					if err := pool.Do(radix.Cmd(nil, "SET", "foo", "bar")); err != nil {
```
nussjustin: Maybe replace this with a call to `radixGetSet`?

mediocregopher: The problem is that `radixGetSet` takes a Conn, not a Client, cause I was worried about the overhead of converting the types to interfaces.

But now that I think about it, I could pre-convert them before the bench starts, and that shouldn't incur any overhead...

bench/bench_test.go
```diff
@@ -22,20 +22,41 @@ func newRedigo() redigo.Conn {
...
+}
+
 func BenchmarkSerialGetSet(b *B) {
 	b.Run("radix", func(b *B) {
 		rad, err := radix.Dial("tcp", "127.0.0.1:6379")
 		if err != nil {
 			b.Fatal(err)
 		}
 		defer rad.Close()
+		client := radix.Client(rad)
```
nussjustin: Maybe add a small comment why this is here, so this doesn't get "optimized" accidentally? Something like

```suggestion
		// avoid overhead of converting from radix.Conn to radix.Client on each loop iteration
		client := radix.Client(rad)
```

bench/bench_test.go
```diff
@@ -187,157 +176,90 @@ func BenchmarkParallelGetSet(b *B) {
...
 
-		b.Run("default", func(b *B) {
-			pool, err := radix.NewPool("tcp", "127.0.0.1:6379", poolSize, radix.PoolPipelineConcurrency(poolSize))
-			if err != nil {
-				b.Fatal(err)
-			}
-			defer pool.Close()
-			if err := waitFull(pool, poolSize); err != nil {
-				b.Fatal(err)
+				client := radix.Client(pool)
```
nussjustin: Same as above

```suggestion
		// avoid overhead of boxing the pool on each loop iteration
		client := radix.Client(rad)
```

bench/bench_test.go
```diff
@@ -107,13 +113,10 @@ func BenchmarkSerialGetSetLargeArgs(b *B) {
 			b.Fatal(err)
 		}
 		defer rad.Close()
+		client := radix.Client(rad)
```
nussjustin: dito

```suggestion
		// avoid overhead of converting from radix.Conn to radix.Client on each loop iteration
		client := radix.Client(rad)
```

mediocregopher:
Lol, I didn't know the suggestion feature had a commit button on it, that's pretty neat! I'll merge once the tests pass.

PR #82, opened 2019-03-15
Optimize clusterSlotKeys brute forcing in tests

nussjustin:
Reduces initialization time for tests by over 200ms on average on my machine, giving a roughly 10x improvement (~239.87ms -> ~22.92ms). This helps when running test on each change.

### DIFF
--- a/cluster_test.go
+++ b/cluster_test.go
@@ -12,7 +12,8 @@ import (
 // second on my laptop, so whatevs.
 var clusterSlotKeys = func() [numSlots]string {
 	var a [numSlots]string
-	for {
+	var found int
+	for found < len(a) {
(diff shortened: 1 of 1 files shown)


cluster_test.go
```diff
@@ -22,22 +23,12 @@ var clusterSlotKeys = func() [numSlots]string {
 			ksmall := k[:i]
 			if a[ClusterSlot(ksmall)] == "" {
 				a[ClusterSlot(ksmall)] = string(ksmall)
+				found++
 				break
 			}
 		}
-
-		var notFull bool
-		for _, k := range a {
```
nussjustin: Turns out iterating over up to 16k strings takes some time when you do this 20,000 times (which is the average number of iterations for the outer loop on my machine)

Issue #91, opened 2019-04-23
bufio.NewReader(r) use memery too mach


qq5272689:
i work with the sentinel module
it's svg https://github.com/qq5272689/test/blob/master/heap_alloc_space.svg

nussjustin:
"too big" is not really helpful.

From what I can guess, your problem are the allocations inside `PubSub`, which come mostly from the call to `bufio.NewReader` inside `resp2.RawMessage.UnmarshalInto`.

In that case, you're right that the code allocates too much and should be optimized to allocate less.

After a quick look I see 2 problems here:

1. The `*bufio.Reader` in `UnmarshalInto` is always created with the default buffer size of 4096 bytes even though we already now the exact number of bytes that are available. For cases where `len(rm) < 4096` we could use `bufio.NewReaderSize` to avoid allocating to much for small messages.
2. The `spin` method of the `pubSubConn` always reads the raw RESP data into a `RawMessage` and after that immediately unmarshals it into a `PubSubMessage`. I don't unterstand why the data is read into a `RawMessage` first. @mediocregopher Any idea?

mediocregopher:
The data is read into a RawMessage first because responses from
PING/(P)(UN)SUBSCRIBE commands also come in through there. So all data is
read into a RawMessage first to be able to determine if it is a
PubSubMessage or not, and discard it if not.

An alternative way to accomplish this, since we don't actually care about
the response from commands, would be to unmarshal directly into
PubSubMessage, and have PubSubMessage.UnmarshalRESP return a specific error
(like `errNotPubSubMessage`) if the data is not a PubSubMessage. The only
caveat would be that the unmarshal method would need to be sure to read the
full response even if the response is not the expected one.

qq5272689:
Thank you so much.

mediocregopher:
I made a further improvement in the last couple of commits, knocking off another bunch of allocations (this is on arm, so exact numbers may vary, and my time/op results were so inconsistent they can probably be ignored):

```
name      old time/op    new time/op    delta
PubSub-4     264µs ±52%     243µs ±52%     ~     (p=0.383 n=20+20)

name      old alloc/op   new alloc/op   delta
PubSub-4      371B ± 1%      160B ± 0%  -56.85%  (p=0.000 n=19+16)

name      old allocs/op  new allocs/op  delta
PubSub-4      10.0 ± 0%       5.0 ± 0%  -50.00%  (p=0.000 n=20+20)
```

PR #92, opened 2019-04-24
Reduce total memory allocated and number of allocations in PubSub processing

nussjustin:
Also includes a new benchmark for simple PubSub

```
name      old time/op    new time/op    delta
PubSub-8    51.9µs ±10%    46.0µs ±14%  -11.24%  (p=0.002 n=10+10)

name      old alloc/op   new alloc/op   delta
PubSub-8    4.73kB ± 0%    0.37kB ± 0%  -92.18%  (p=0.000 n=10+10)

name      old allocs/op  new allocs/op  delta
PubSub-8      17.0 ± 0%      10.0 ± 0%  -41.18%  (p=0.000 n=10+10)
```

Updates #91

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -2,7 +2,6 @@ package radix
 
 import (
 	"bufio"
-	"bytes"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


mediocregopher:
Thanks justin! We could probably save some more time by not using RawMessage at all, but that can be done another time, since this already saves so much in allocations as it is.

PR #95, opened 2019-05-03
Avoid allocating a string for each struct field when unmarshalling

nussjustin:
This reuses the allocated memory for the struct field by switchting from `BulkString` to `BulkStringBytes` and using the `map[string([]byte)]` optimization introduced in Go 1.3.

```
name                       old time/op    new time/op    delta
AnyUnmarshalRESP/Struct-8    1.68µs ± 1%    1.35µs ± 1%  -19.73%  (p=0.000 n=9+8)

name                       old alloc/op   new alloc/op   delta
AnyUnmarshalRESP/Struct-8      112B ± 0%       96B ± 0%  -14.29%  (p=0.000 n=10+10)

name                       old allocs/op  new allocs/op  delta
AnyUnmarshalRESP/Struct-8      7.00 ± 0%      4.00 ± 0%  -42.86%  (p=0.000 n=10+10)
```

### DIFF
--- a/resp/resp2/bench_test.go
+++ b/resp/resp2/bench_test.go
@@ -38,3 +38,28 @@ func BenchmarkIntUnmarshalRESP(b *testing.B) {
 		})
 	}
 }
+
+func BenchmarkAnyUnmarshalRESP(b *testing.B) {
+	b.Run("Struct", func(b *testing.B) {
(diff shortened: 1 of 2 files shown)


mediocregopher:
Nice, thanks!!

PR #96, opened 2019-05-04
Reuse reflect.Value instances when unmarshalling maps

nussjustin:
Unmarshalling maps currently requires creating a new instance of both the key and value types for each key and value, meaning there are at least two allocations per key-value pair, even if the types themselves wouldn't require allocations (e.g. numbers).

For non-pointer values that don't contain pointers or where the pointed to values can't be changed (strings), we can reuse the `reflect.Value` instances to avoid allocating new `reflect.Value`s for each value instance that we need.

This change adds a check for these kind of values (basically all built in numeric types, booleans and strings) and reuses the `reflect.Value` instances for these if possible.

With this for a map of size `n`, where `n > 0` and both key and value can share their `reflect.Value` instance, the number of allocations is reduced by `2*n - 2`.

```
name                    old time/op    new time/op    delta
AnyUnmarshalRESP/Map-8    2.07µs ± 1%    1.76µs ± 7%  -14.69%  (p=0.000 n=20+18)

name                    old alloc/op   new alloc/op   delta
AnyUnmarshalRESP/Map-8      504B ± 0%      408B ± 0%  -19.05%  (p=0.000 n=20+20)

name                    old allocs/op  new allocs/op  delta
AnyUnmarshalRESP/Map-8      18.0 ± 0%      12.0 ± 0%  -33.33%  (p=0.000 n=20+20)
```

### DIFF
--- a/resp/resp2/bench_test.go
+++ b/resp/resp2/bench_test.go
@@ -40,6 +40,29 @@ func BenchmarkIntUnmarshalRESP(b *testing.B) {
 }
 
 func BenchmarkAnyUnmarshalRESP(b *testing.B) {
+	b.Run("Map", func(b *testing.B) {
+		b.ReportAllocs()
+
+		const input = "*8\r\n" +
+			"$3\r\nFoo\r\n" + "$1\r\n1\r\n" +
+			"$3\r\nBAZ\r\n" + "$2\r\n22\r\n" +
+			"$3\r\nBoz\r\n" + "$4\r\n4444\r\n" +
+			"$3\r\nBiz\r\n" + "$8\r\n88888888\r\n"
+
+		var sr strings.Reader
+		br := bufio.NewReader(&sr)
+
+		for i := 0; i < b.N; i++ {
+			sr.Reset(input)
+			br.Reset(&sr)
+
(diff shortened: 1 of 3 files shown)


nussjustin:
@mediocregopher Thought about this yesterday and did a few tests locally and couldn't find a case where this optimization would cause a bug/regression, but since it's quite a bit subtle I wouldn't be surprised if I overlooked something.

mediocregopher:
I like it! It's very clever lol. I feel like it's not a use-case of reflect that the authors explicitly support but I doubt they'd ever break it either.

nussjustin:
> I feel like it's not a use-case of reflect that the authors explicitly support but I doubt they'd ever break it either.

Well it's basically a (mis)use of value semantics and the immutability of strings, so I don't see how they could ever break it, but I've never seen anyone else do this. I'm still surprised this works, even though I know why it works. And the reduction in allocation is really nice (in my current use case maps can have over 100 entries, which means >200 allocs less per map and I have thousands of them)

mediocregopher:
> which means >200 allocs less per map and I have thousands of them)

:O !!!

PR #99, opened 2019-05-21
[WIP] Add a feature to get metric by giving Trace struct

kimtree:
Hi,
This is a way better, more human readable approach. User can create trace functions like below.
```
ct := radix.ClusterTrace{
	TopoChange: func() {
		fmt.Println("topology changed")
	},
	MovedRespond: func() {
		fmt.Println("moved!")
	},
}
pt := radix.PoolTrace{
	ConnectDone: func(addr string, err error) {
		if err != nil {
			fmt.Println("err")
		} else {
			fmt.Println(addr)
		}
	},
}
```

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -312,7 +312,9 @@ func (c *Cluster) sync(p Client) error {
 	}
 
 	if reflect.DeepEqual(c.topo.Map(), tt.Map()) {
-		go c.co.ct.TopoChange()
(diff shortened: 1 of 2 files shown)


kimtree:
@mediocregopher Thanks for the comments! This PR is just for POC so I will upload new draft PR for discussing more about this improvement.

Issue #93, opened 2019-04-25
Questions about Sync() in Cluster


kimtree:
Hi, I got more questions while I'm benchmarking your library.

I was trying to get a count for calling Sync() function while running a huge amount of insertion command. This investigation helps me to know how much time needed for cluster changing and how does this affect overall performance.
For more detail, I modified the like below
``` cluster.go
type Cluster struct {
        ~ blah blah ~
        CounterSync uint32 // This is what I added to get a count
}
func (c *Cluster) Sync() error {
        atomic.AddUint32(&c.CounterSync, 1) // This is what I added to get a count
        p, err := c.pool("")
        if err != nil {
                return err
        }
        c.syncDedupe.do(func() {
                err = c.sync(p)
        })
        return err
}
```
And also, I added some sort of lines for measure time like below
```
func (c *Cluster) sync(p Client) error {
        start := time.Now()

        ~  blah blah ~

        fmt.Printf("\n%d", time.Since(start).Nanoseconds())
        return nil
}
```
And this is my benchmark code (40 goroutines)
```
func BenchmarkTestQPS(b *testing.B) {
	b.Run("zincrby", func(b *testing.B) {
		client := radixClusterWithAuthPass(20) // This function will return Cluster which has 20 connections in each pool

		b.ResetTimer()
		b.RunParallel(func(pb *testing.PB) {
			for pb.Next() {
				client.Do(radix.Cmd(nil, "ZINCRBY", "test_key", "1", "test_data"))
			}
		})
		b.StopTimer()
                 fmt.Println("Value Of CounterSync:  ", client.CounterSync)
	})
}
```
And the result seemed like this
```
2883036
592498
605760
586586
756704
661927
535979
506501
580632
Value Of CounterSync:  3
```
This seems that sync() function called 9 times while the counter shows only 3. Do you have any idea what's happening here?

mediocregopher:
Hi @kimtree! That really is odd, I can't imagine why that would be happening. I tried to replicate your results and was not able to. I more or less made the same changes you did, you can see my diff 
[here](https://github.com/mediocregopher/radix/files/3117690/qps.diff.txt), and this is the output I get:

```
>>> go test -v -run=xxx -bench=QPS .       
sync took: 8.897832ms                                              
Value Of CounterSync:   1                                 
goos: linux                                                  
goarch: arm64                                                             
pkg: github.com/mediocregopher/radix/v3                                           
[...]

PR #100, opened 2019-05-22
Add a feature to get informations in Cluster/Pool by using Trace struct

kimtree:
Hi, I've uploaded new PR and improve and clean up the codes little bit. I'm not sure I fixed this PR clearly what you mentioned before.

This PR allows people to get callbacks for a specific time of the event (which is described in trace package). I made simple Trace struct and callback events for Cluster, Pool respectively for the first version.

Could you please check whether the callback events propagate at the right time of the event and review this PR?

Can be improved...
  - [x] Make static reason values (REFILL, BUFFERED, etc) to a constant value
  - [x] Find a better way to find added, removed cluster nodes when topology changed
  - Give more detailed descriptions for the trace package (Sorry for my poor English in docstring)
  - Add TCs

Any suggestions welcome!

ref: https://github.com/mediocregopher/radix/issues/93

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -134,7 +134,6 @@ func NewCluster(clusterAddrs []string, opts ...ClusterOpt) (*Cluster, error) {
 	defaultClusterOpts := []ClusterOpt{
 		ClusterPoolFunc(DefaultClientFunc),
 		ClusterSyncEvery(5 * time.Second),
-		ClusterWithTrace(trace.ClusterTrace{}),
 	}
 
 	for _, opt := range append(defaultClusterOpts, opts...) {
--- a/cluster.go
+++ b/cluster.go
@@ -460,7 +460,7 @@ func (c *Cluster) Do(a Action) error {
 
 func (c *Cluster) traceRedirected(addr, key string, moved, ask bool, attempts int) {
 	if c.co.ct.Redirected != nil {
-		c.co.ct.Redirected(trace.ClusterRedirected{Addr: addr, Key: key, Moved: moved, Ask: ask, RemainRedirectCount: attempts})
+		c.co.ct.Redirected(trace.ClusterRedirected{Addr: addr, Key: key, Moved: moved, Ask: ask, RedirectCount: doAttempts - attempts})
 	}
 }
 
--- a/cluster.go
+++ b/cluster.go
@@ -440,9 +440,9 @@ func (c *Cluster) Do(a Action) error {
 	return c.doInner(a, addr, key, false, doAttempts)
 }
 
-func (c *Cluster) traceGotResponse(addr, key string, attempts int, err error) {
-	if c.co.ct.GotResponse != nil {
-		c.co.ct.GotResponse(trace.ClusterGotResponse{Addr: addr, Key: key, RetryCount: attempts, Err: err})
+func (c *Cluster) traceRedirected(addr, key string, moved, ask bool, attempts int) {
+	if c.co.ct.Redirected != nil {
+		c.co.ct.Redirected(trace.ClusterRedirected{Addr: addr, Key: key, Moved: moved, Ask: ask, RemainRedirectCount: attempts})
 	}
 }
 
@@ -465,7 +465,6 @@ func (c *Cluster) doInner(a Action, addr, key string, ask bool, attempts int) er
 	}
 
 	err = p.Do(thisA)
-	c.traceGotResponse(addr, key, attempts, err)
 	if err == nil {
 		return nil
 	}
@@ -477,6 +476,7 @@ func (c *Cluster) doInner(a Action, addr, key string, ask bool, attempts int) er
 	if !moved && !ask {
 		return err
 	}
+	c.traceRedirected(addr, key, moved, ask, attempts)
 
 	// if we get an ASK there's no need to do a sync quite yet, we can continue
 	// normally. But MOVED always prompts a sync. In the section after this one
--- a/cluster.go
+++ b/cluster.go
@@ -461,18 +461,15 @@ func (c *Cluster) doInner(a Action, addr, key string, ask bool, attempts int) er
 	thisA := a
 	if ask {
 		thisA = WithConn(key, func(conn Conn) error {
-			err = askConn{conn}.Do(a)
-			c.traceGotResponse(addr, key, attempts, err)
-			return err
+			return askConn{conn}.Do(a)
 		})
 	}
 
-	if err = p.Do(thisA); err == nil {
-		c.traceGotResponse(addr, key, attempts, nil)
+	err = p.Do(thisA)
+	c.traceGotResponse(addr, key, attempts, err)
+	if err == nil {
 		return nil
 	}
-
-	c.traceGotResponse(addr, key, attempts, err)
 
 	// if the error was a MOVED or ASK we can potentially retry
 	msg := err.Error()
--- a/cluster.go
+++ b/cluster.go
@@ -296,25 +296,25 @@ func (c *Cluster) traceTopoChanged(prevTopo ClusterTopo, newTopo ClusterTopo) {
 	if c.co.ct.TopoChanged != nil {
 		var addedNodes []trace.ClusterNodeInfo
 		var removedNodes []trace.ClusterNodeInfo
+		var changedNodes []trace.ClusterNodeInfo
 
 		prevTopoMap := prevTopo.Map()
+		newTopoMap := newTopo.Map()
 
-		for addr, newNode := range newTopo.Map() {
+		for addr, newNode := range newTopoMap {
 			if prevNode, ok := prevTopoMap[addr]; ok {
-				if !reflect.DeepEqual(newNode, prevNode) {
-					// Previous Node changed its information
-					removedNodes = append(removedNodes, trace.ClusterNodeInfo{
-						Addr:      prevNode.Addr,
-						Slots:     prevNode.Slots,
-						IsPrimary: prevNode.SecondaryOfAddr == "",
-					})
-					addedNodes = append(addedNodes, trace.ClusterNodeInfo{
+				// Check whether two nodes which have the same address changed its value or not
+				if !reflect.DeepEqual(prevNode, newNode) {
+					changedNodes = append(changedNodes, trace.ClusterNodeInfo{
 						Addr:      newNode.Addr,
 						Slots:     newNode.Slots,
 						IsPrimary: newNode.SecondaryOfAddr == "",
 					})
 				}
+				// No need to handle this address for finding removed nodes
+				delete(prevTopoMap, addr)
 			} else {
+				// The node's address not found from prevTopo is newly added node
 				addedNodes = append(addedNodes, trace.ClusterNodeInfo{
 					Addr:      newNode.Addr,
 					Slots:     newNode.Slots,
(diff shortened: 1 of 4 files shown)


cluster.go
```diff
@@ -281,6 +293,34 @@ func (c *Cluster) Sync() error {
...
+		}
+	}
+	return false
+}
+
+func (c *Cluster) traceTopoChanged(prevTopo ClusterTopo, newTopo ClusterTopo) {
+	if c.co.ct.TopoChanged != nil {
+		var Added []trace.ClusterNodeInfo
+		var Removed []trace.ClusterNodeInfo
+		for _, t := range newTopo.Map() {
```
nussjustin: `Map` returns a map keyed by the node address. You could just check which keys are in the old map that are missing in the new map and vice versa. For keys that exists in both you could compare IDs and slots and track changes separate from added and removed nodes.

cluster.go
```diff
@@ -281,6 +293,34 @@ func (c *Cluster) Sync() error {
...
+		if reflect.DeepEqual(t, node) {
+			return true
+		}
+	}
+	return false
+}
+
+func (c *Cluster) traceTopoChanged(prevTopo ClusterTopo, newTopo ClusterTopo) {
+	if c.co.ct.TopoChanged != nil {
+		var Added []trace.ClusterNodeInfo
```
nussjustin: Local variables should be lower case in Go. Same for all the other places.

cluster.go
```diff
@@ -406,14 +454,19 @@ func (c *Cluster) doInner(a Action, addr, key string, ask bool, attempts int) er
 	thisA := a
 	if ask {
 		thisA = WithConn(key, func(conn Conn) error {
-			return askConn{conn}.Do(a)
+			err = askConn{conn}.Do(a)
+			c.traceGotResponse(addr, key, attempts, err)
```
nussjustin: This will call `traceGotResponse` multiple times won't it? Once here and than below on line 464 or 468. In that case, just remove this.

cluster.go
```diff
@@ -406,14 +454,19 @@ func (c *Cluster) doInner(a Action, addr, key string, ask bool, attempts int) er
...
+			return err
 		})
 	}
 
 	if err = p.Do(thisA); err == nil {
+		c.traceGotResponse(addr, key, attempts, nil)
```
nussjustin: You could instead to 

```go
err = p.Do(thisA)
c.traceGotResponse(addr, key, attempts, err)
if err != nil {
    return nil
}
```

and remove the call to `traceGotResponse` on line 468.

pool.go
```diff
@@ -25,10 +26,12 @@ type ioErrConn struct {
 	// level error, e.g. a timeout, disconnect, etc... Close is automatically
 	// called on the client when it encounters a critical network error
 	lastIOErr error
+
+	ConnectTime time.Duration
```
nussjustin: The struct is not exported, so it does not make sense to export this.

pool.go
```diff
@@ -334,13 +349,44 @@ func (p *Pool) err(err error) {
...
 // this must always be called with p.l unlocked
 func (p *Pool) newConn(errIfFull bool) (*ioErrConn, error) {
+	start := time.Now()
 	c, err := p.opts.cf(p.network, p.addr)
+	elapsed := time.Since(start)
 	if err != nil {
 		return nil, err
 	}
-	ioc := newIOErrConn(c)
+	ioc := newIOErrConn(c, elapsed)
```
nussjustin: Instead of tracking the time inside the `*ioErrConn` and tracing the event in the caller, why not just trace the event here? If it's for the `reason` phrase, you could just pass it as an argument to `newConn` (which I can also imagine being nice when debugging).

Also that way we could trace the time it took (could be interesting for timeouts).

pool.go
```diff
@@ -334,13 +349,44 @@ func (p *Pool) err(err error) {
 	}
 }
 
+func (p *Pool) traceConnectDone(conn *ioErrConn, reason string, err error) {
+	if p.opts.pt.ConnectDone != nil {
+		var connectTime time.Duration
+		if conn != nil {
+			connectTime = conn.ConnectTime
+		} else {
+			connectTime = 0
```
nussjustin: This is unnecessary since `connectTime` is already 0.

trace/trace.go
```diff
@@ -0,0 +1,68 @@
...
+	Slots     [][2]uint16
+	IsPrimary bool
+}
+
+type ClusterTopoChanged struct {
+	Added   []ClusterNodeInfo
+	Removed []ClusterNodeInfo
+}
+
+type ClusterGotResponse struct {
```
nussjustin: I don't really like this one. Instead of tracing just how many tries a command needed, we could just trace `ASK` and `MOVED` directly. What do you think?

cluster.go
```diff
@@ -123,6 +134,7 @@ func NewCluster(clusterAddrs []string, opts ...ClusterOpt) (*Cluster, error) {
 	defaultClusterOpts := []ClusterOpt{
 		ClusterPoolFunc(DefaultClientFunc),
 		ClusterSyncEvery(5 * time.Second),
+		ClusterWithTrace(trace.ClusterTrace{}),
```
nussjustin: This is unnecessary, since a new Cluster will already have `ct` set to an empty `ClusterTrace`.

pool.go
```diff
@@ -261,6 +273,7 @@ func NewPool(network, addr string, size int, opts ...PoolOpt) (*Pool, error) {
 		PoolPipelineConcurrency(size),
 		// NOTE if 150us is changed the benchmarks need to be updated too
 		PoolPipelineWindow(150*time.Microsecond, 0),
+		PoolWithTrace(trace.PoolTrace{}),
```
nussjustin: Same as with the `WithClusterTrace` call: This is unnecessary, since a new Pool will already have `pt` set to an empty `PoolTrace`.

cluster.go
kimtree: Yes, I'll change it.

cluster.go
kimtree: That's a good idea. I'll change it to iterate newTopo's `Map` and check whether it was added or changed.

cluster.go
kimtree: I misunderstood. I should remove this line. Thanks.

cluster.go
kimtree: we could change this line.
```golang
if err == nil {  // changed condition
    return nil
}
```

pool.go
kimtree: I'll fix it.

pool.go
kimtree: Yes, that was the reason. I'll change it not to hold tracked time in newIOErrConn and make newConn function handle reason.

pool.go
kimtree: I'll remove this variable.

trace/trace.go
kimtree: What I was trying to get is just for `MOVED` command (which is a great monitoring metric that can be used for checking cluster topology has been changed). I can change this to handle `MOVED` command only If you think `ClusterGotResponse` is too broad.

cluster.go
kimtree: I'll fix it.

pool.go
kimtree: I'll fix it.

cluster.go
```diff
@@ -281,6 +292,41 @@ func (c *Cluster) Sync() error {
...
+			} else {
+				addedNodes = append(addedNodes, trace.ClusterNodeInfo{
+					Addr:      newNode.Addr,
+					Slots:     newNode.Slots,
+					IsPrimary: newNode.SecondaryOfAddr == "",
+				})
+			}
+		}
+
+		c.co.ct.TopoChanged(trace.ClusterTopoChanged{Added: addedNodes, Removed: removedNodes})
```
nussjustin: You forgot to check for removed nodes. Right now you're only tracking added and changed nodes.

trace/trace.go
```diff
@@ -0,0 +1,68 @@
...
+type PoolConnInfo struct {
+	PoolSize   int
+	BufferSize int
+	AvailCount int
+}
+
+type PoolConnectDone struct {
+	PoolInfo
+	PoolConnInfo
+	Reason      string
```
mediocregopher: These `Reason` fields are going to need to be enumerated as consts in this package, as their own type:

```
type PoolConnectReason string

const (
    PoolConnectReasonInitialize PoolConnectReason = "initialize"
    PoolConnectReasonRefill PoolConnectReason = "refill"
    ....
)
```

You can leave the "TEST" reasons as-is, cause tests don't matter.

trace/trace.go
mediocregopher: I think a more usable version of this might look like:

```go
type ClusterRedirect struct {
    Addr string
    Key string // if possible
    Moved, Ask bool
    RedirectCount int // basically RetryCount
}
```

I don't think it's necessary to be tracing on _every_ response, only the redirect ones.

mediocregopher: So basically, nearly the same as you have now, but moving it to (here)[https://github.com/mediocregopher/radix/pull/100/files#diff-5012ac478fddb03fa409c41a16a72b0dL424}

kimtree: Agree. Tracing the redirect response makes sense to me as well. Thanks. I'll fix it.

cluster.go
kimtree: Oh, my mistake! I'll check the added, removed, changed nodes.

trace/trace.go
kimtree: That's the idea what I want to know. Managing reasons in a separate package are fine.

trace/trace.go
kimtree: For the `RedirectCount`, It seems that radix library only handles **remain** retry counts for the response. Should I change it to `RemainRetryCount` or fix the codes to keep actual retried count?

mediocregopher: I think it'd be fine to pass `doAttempts - attempts` as the `RedirectCount` value.

kimtree: Okay, I'll fix it.

kimtree:
@mediocregopher @nussjustin Hi Guys, Is there anything else I can do for this PR? I need this PR to be merged and released with the new version tag cause I'm planning to migrate our Redis library to Radix ASAP.

mediocregopher:
I'm ok with merging the code as it is, I think you did a great job getting the basics fleshed out. If I could ask you for one last thing: I think the `trace` package should have a package comment that states very explicitly that the package is not stable yet and is subject to change. This would go against go's backwards compatibility standards for modules, but since this is a subpackage that most people won't use I think it'll be ok. And since there's so many new public types it's really unlikely nothing is going to change.

kimtree:
@mediocregopher Thanks for reviewing this. One last change is changing PoolInfo to PoolHostInfo that store Network value as well and change previous PoolConnInfo to PoolInfo to store pool's information. And also, I added disclaimer note on the top of the trace package file.

mediocregopher:
Thanks @kimtree! I'll probably go in soon and update the docs in trace, and might change some of the types slightly.

kimtree:
@mediocregopher Thank you so much! And do you a have plan to release new version soon? I’m working on applying this trace feature in our project, so I would appreciate If I can have new version tag!

mediocregopher:
Not a problem, thanks so much for implementing it! I hope to tag it soon,
once I flesh out the docs a bit, but I won't make any promises about a
timeline. In the meantime you can use the commit in your `go.mod` by
calling `go get github.com/mediocregopher/radix/v3@85e7773`, and that'll
get you the trace code for now.

kimtree:
Thanks for the info! I’ll try the command in the meantime~ :smile:

Issue #103, opened 2019-05-23
Add support for TLS connections


Resonance1584:
We are using AWS Elasticache for Redis which adds an optional TLS layer to the connection. https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/in-transit-encryption.html

This feature is required in order to use AUTH with Elasticache.

It is a relatively minor change to add a TLS Dial option to radix, but I do understand if you'd rather keep this library simpler.

mediocregopher:
Merged, thanks for the PR! :)

PR #104, opened 2019-05-23
Add DialUseTLS Option

Resonance1584:
Suggested implementation for #103

The actual implementation here is quite simple, the complex part is terminating TLS in a valid test case. I've tested this manually against AWS ElastiCache for Redis In-Transit Encryption (TLS).

### DIFF
--- a/radix.go
+++ b/radix.go
@@ -131,6 +131,7 @@ package radix
 
 import (
 	"bufio"
+	"crypto/tls"
 	"errors"
 	"net"
 	"net/url"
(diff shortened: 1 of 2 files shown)


mediocregopher:
hi @Resonance1584 , `DialUseTLS` looks good to me, thanks for adding it! It looks like the tests are failing because you're using `strings.ReplaceAll` which wasn't added until go 1.12, but we always support the most recent two versions.

Resonance1584:
@mediocregopher great, I've pushed an update to remove `strings.ReplaceAll`

PR #105, opened 2019-05-25
Remove a wrong comment on PubSubMessage

nussjustin:
The [comment on MarshalRESP](https://github.com/mediocregopher/radix/blob/master/pubsub.go#L24-L25) says that a message with a pattern will be assumed to be a PMESSAGE, but there is [no such logic](https://github.com/mediocregopher/radix/blob/master/pubsub.go#L34-L43).

### DIFF
--- a/pubsub.go
+++ b/pubsub.go
@@ -21,8 +21,7 @@ type PubSubMessage struct {
 	Message []byte
 }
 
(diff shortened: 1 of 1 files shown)


mediocregopher:
Woops, yeah there used to be a bunch of gob message handling stuff that got gotten rid of. Thanks!

Issue #107, opened 2019-05-29
Possible wrong value of AvailCount in PoolTrace function


kimtree:
Hi Guys, 
I found a possible wrong value in PoolTrace function. The PoolConnectDone function provides `AvailCount` by using len(p.pool), but newly created connection is not added to the pool at this moment (when PoolConnectDone function called). So, AvailCount is not the same as I expected when I check value right after the creation of a new cluster(expect: pool size, actual: pool size - 1).
ref: https://github.com/mediocregopher/radix/blob/a4b003e8968c1faef8f031557bcb25a7cdd70afd/pool.go#L370-L378

Do you guys have any idea to get Pool's current available connection count? IMO, adding interval function in PoolTrace and make it run inside of Pool's atIntervalDo() and get available connection count can be a simple idea. I'll upload a PR for this If you guys agree.

mediocregopher:
Hi @kimtree, sorry it's been taking me so long to respond. It is a bit annoying that using `ConnectDone` won't ever give the correct pool size. At the same time, I don't think using `atIntervalDo` is a good idea, since we shouldn't add a running go-routine for the sole purpose of tracing imo.

One solution, which would also solve another problem, would be to add a `DoCompleted` method to `PoolTrace`, which would get invoked _after_ the `put` has been called (or after `p.pipeliner.Do` has returned). `PoolDoCompleted` would have all the normal attributes of the others, as well as an `DoTime` which would be the total time it took to complete the `Do`, from start to end.

I think this solution will allow you to retrieve a correct pool count frequently as well as provide for other use-cases as well. What do you think? Also, keep in mind that if you need to get the available connection count on-demand there's always the `NumAvailConns` method on `Pool`.

kimtree:
Never mind for responding late 😄  I think the `PoolDoCompleted` would be great and provide completion time for `Do` function is a good metric value for the users.

I remember that there's a `NumAvailConns` method on `Pool`, but I couldn't access it from the Cluster. And also, making another routine for getting `NumAvailConns` periodically and send it to Grafana or something can be an extra logic for me (compare to use `PoolTrace` only)

mediocregopher:
Merged, thanks again! :)

PR #108, opened 2019-06-03
Add DoComplete trace for Pool

kimtree:
Ref Issue: https://github.com/mediocregopher/radix/issues/107

I've added new trace type for tracking the completion of `Do` command. This enables people to get elapsed time for executing a command and the information of connection counts in `Pool`.


pool.go
```diff
@@ -567,7 +571,21 @@ func (p *Pool) Do(a Action) error {
 	}
 	defer p.put(c)
 
-	return c.Do(a)
+	startTime := time.Now()
+	err = c.Do(a)
+	p.traceDoCompleted(time.Since(startTime), err)
```
mediocregopher: The way this is implemented right now, the trace does not include the `get` or the `put` calls, and since it happens before the `put` it also won't ever show the Pool as being "full". I think `startTime` should initialize before `get`, and `put` should be taken out of the defer statement so `traceDoCompleted` can happen after it.

kimtree: I changed to initialize `startTime` at the first line of the `Do` function and call `traceDoComplete` right after `put` connection in `Pool`.

kimtree: Oh, I think I'm misunderstood. Did you mean to add a trace for `get`, `put` and also `DoComplete` events? How about making `PoolChange` or some trace for detecting `get`, `put` event in `Pool` and separate `DoComplete` trace?

mediocregopher: What you did here is what I had meant, no worries. I don't think it'd be useful to have `get` and `put` traces right now, I think `DoCompleted` is fine.

kimtree: @mediocregopher Sounds good, Thanks!

PR #111, opened 2019-06-07
Add InitCompleted trace and get avail count for specified events

kimtree:
Hi, I've changed for getting (quite) correct value for available connection count in Pool. I added a new trace function named `InitCompleted` which invokes right after `Pool` fills all connections when initializing.

And then, We can get the correct value for available connection count for three cases. First, `initCompleted` will return `AvailCount` which is the same size as the `Pool`. Second, `DoComplete` will callback right after returning the connection into the `Pool` so we could get the exact value. Third, `ConnClosed`. This also callback after we get the connection to be closed, therefore we could get exact connection count as well.

### DIFF
--- a/pool.go
+++ b/pool.go
@@ -295,6 +295,7 @@ func NewPool(network, addr string, size int, opts ...PoolOpt) (*Pool, error) {
 
 	p.wg.Add(1)
 	go func() {
+		startTime := time.Now()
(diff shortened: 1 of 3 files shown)


mediocregopher:
Great PR, thanks for adding this :)

Issue #109, opened 2019-06-05
Is there a way to handle error gently?


kimtree:
Hi, I'm trying to handle the error by types. I made some codes to invoke timeout error(by giving extremely short timeout) when creating a connection in Pool. But, I couldn't detect timeout error by checking error type like `err.(net.Error)`. I think the line below is the cause.

https://github.com/mediocregopher/radix/blob/master/cluster.go#L359
How about making every detectable error as constant that people can handle error more easily?

And Is it possible to distinguish error like long execution time when client.Do() called or timeout while creating a connection in Pool?

```golang
customConnFunc := func(network, addr string) (radix.Conn, error) {
	return radix.Dial(network, addr,
		radix.DialReadTimeout(50 * time.Millisecond),
		radix.DialConnectTimeout(100 * time.Millisecond),
		radix.DialAuthPass("*********"),
	)
}	

poolFunc := func(network, addr string) (radix.Client, error) {
	return radix.NewPool(network, addr, 1, radix.PoolConnFunc(customConnFunc))
}

c, err := radix.NewCluster(ClusterHostsWithAuthPass, radix.ClusterPoolFunc(poolFunc))
if err != nil {
	fmt.Println(err)
	if netErr, ok := err.(net.Error); ok && netErr.Timeout() {
		fmt.Println("Timeout")
	} else if opError, ok := err.(*net.OpError); ok {
		if opError.Op == "dial" {
			fmt.Println("Unknown host")
		} else if opError.Op == "read" {
			fmt.Println("Connection refused")
		}
	} else {
		fmt.Println("Etc errors")
	}
}
```

mediocregopher:
Hmmmm yeah that is tricky. In go 1.13 they are going to be releasing changes to the `errors` package which will help address exactly this problem. I will be able to change that line to:

```go
fmt.Errorf("error connecting to %s: %w", t.Addr, err)
```

And you'll be able to check it using the `errors.As` function. Obviously those aren't available yet though, so I'm not sure. Maybe you can just do string checking for now, and switch to the new error functions once 1.13 is released?

nussjustin:
@mediocregopher There is a golang.org/x/xerrors package which can be used right now and which will automatically use the stdlib logic on Go 1.13.

mediocregopher:
oooooooooooo ok if I have time later today I'll check that out

mediocregopher:
[...]

PR #113, opened 2019-06-10
switch to use xerrors temporarily

mediocregopher:
go 1.13 will introduce new functions and fuctionality to the errors
package, but older versions of go will not receive these upgrades. To
compensate there has been released an xerrors package, which seemlessly
implements these upgrades. We will keep using xerrors so that older go
versions can make use of these changes. We will do so until go 1.14 is
released, at which point all supported go versions will have the new
error functionality.

(updates #109)

### DIFF
--- a/bench/bench_test.go
+++ b/bench/bench_test.go
@@ -2,11 +2,12 @@ package bench
 
 import (
 	"context"
-	"errors"
 	"runtime"
 	"strings"
 	. "testing"
 	"time"
+
+	errors "golang.org/x/xerrors"
 
 	redigo "github.com/gomodule/redigo/redis"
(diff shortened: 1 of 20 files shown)


templexxx:
@mediocregopher 
xerrors need upgrade becasue Go1.13 errors had break changes.

build will fail with Go1.13:
golang.org/x/xerrors@v0.0.0-20190513163551-3ee3066db522/adaptor_go1_13.go:16:14: undefined: errors.Frame
golang.org/x/xerrors@v0.0.0-20190513163551-3ee3066db522/format_go1_13.go:12:18: undefined: errors.Formatter

PR #120, opened 2019-06-12
sentinel: have NewSentinel extract any AUTH creds out of the given ur…

### LINKED ISSUE #112: passing AUTH password via URI ONLY in NewSentinel makes PersistentPubSub blocks
I assume PersistentPubSub only get password from DialAuthPass. 
If we do "redis://:fancyPassword@127.0.0.1:26379", PersistentPubSub will not use the fancyPassword to sub "switch-master".

mediocregopher:
…l, if the address is given as a url, and use them for all future sentinel conns. Fixes #112.

### DIFF
--- a/radix.go
+++ b/radix.go
@@ -245,6 +245,13 @@ var DefaultConnFunc = func(network, addr string) (Conn, error) {
 	return Dial(network, addr)
 }
 
+func wrapDefaultConnFunc(addr string) ConnFunc {
+	_, opts := parseRedisURL(addr)
+	return func(network, addr string) (Conn, error) {
+		return Dial(network, addr, opts...)
+	}
+}
+
 ////////////////////////////////////////////////////////////////////////////////
 
 type dialOpts struct {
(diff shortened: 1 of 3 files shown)


radix.go
```diff
@@ -348,6 +355,39 @@ var defaultDialOpts = []DialOpt{
...
+		opts = append(opts, DialAuthPass(p))
+	} else if qpw := q.Get("password"); qpw != "" {
+		opts = append(opts, DialAuthPass(qpw))
+	}
+
+	var dbStr string
+	if u.Path != "" && u.Path != "/" {
+		dbStr = u.Path[1:]
+	} else if dbStr = q.Get("db"); dbStr != "" {
+		// cool
```
nussjustin: Nitpick: You could move the assignment into the `else if` and change it to an `else` or even remove the whole `else` by initializing `dbStr` to the result of `q.Get("db")` like this:

```go
dbStr := q.Get("db")
if u.Path != "" && u.Path != "/" {
	dbStr = u.Path[1:]
}
```

mediocregopher: I was being lazy >_< but you're right I should make it nice

mediocregopher:
> LGTM
> 
> /ot Looking at the code reminded me of the new [ACLs](https://redis.io/topics/acl) coming in Redis 6. I think I'll start to play a bit more the the unstable branch again.

Oh geez, I hadn't seen that. Looks like the new Auth won't be _too_ difficult to add support for, but there'll be another DialAuth function which'll be ugly -_-

Issue #53, opened 2018-12-07
Broken govendor/dep.. vendor support


mieczkowski:
@mediocregopher I appreciate that you target to new modules feature in golang, but imports with /v3/ in path broke support for old GOPATH way (try using govendor/dep - it is not working anymore). Please consider some backward compatibility, because some people are not ready for changing dependencies managment system (and don't want to change library, because radix fits perfectly our needs :-) )

yerden:
+1. `dep` is apparently broken.
``` bash
init failed: unable to solve the dependency graph: Solving failure: No versions of github.com/mediocregopher/radix met constraints:
        3.1.0: Could not introduce github.com/mediocregopher/radix@3.1.0 due to multiple problematic subpackages:
        Subpackage github.com/mediocregopher/radix/v3/resp is missing. (Package is required by github.com/mediocregopher/radix.v3@master.)      Subpackage github.com/mediocregopher/radix/v3/resp/resp2 is missing. (Package is required by github.com/mediocregopher/radix.v3@master.)
```

mediocregopher:
#50 already reported this. Unfortunately there's not much I can do at this point, it's too late to change the import paths of the package. It looks like dep has a PR open to fix this here: https://github.com/golang/dep/pull/1963

but the author maybe just had a baby, judging by the comments. I'll leave this issue open, waiting on that one to get closed.

mieczkowski:
@mediocregopher  how about putting `resp` package into te main package? Problematic imports are gone and we have backwards compability ;-)

yerden:
@mieczkowski I think, just as @mediocregopher stated, that would break builds for a lot of dependency packages. Hence, that would also break backwards compability.

awkejiang:
any updates on this issue ?

mediocregopher:
As i mentioned [here](https://github.com/mediocregopher/radix/issues/53#issuecomment-445397343) there's not really anything to be done. Once again, someone could try contributing to dep to get that PR through, that would be the most effective fix.

cep21:
You can support both if you make a `/v3` directory and move all the code to that directory.  It will support both modules and dep.  Will you accept a pull request for this?

cep21:
I made an example of what it would look like to support dep and modules at the same time as #128 It's just a reference PR to see what it looks like.

mediocregopher:
Hi @cep21, thanks so much for bringing this up! I'm going to continue the convo here rather than the PR.

[...]

PR #128, opened 2019-06-20
Example of supporting dep and modules

cep21:
An example of what it would look like to support dep and modules at the same time.


cep21:
This is mentioned as an alternative under https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher "Major subdirectory" approach.  This is understandable by older versions of go, including dep.  Go modules are also aware of this approach and will look in the /v3 directory for the right files.

On #53 you had a concern
> only way that would work is if I copied resp and its sub-package into the main one

This keeps resp in it's own package.

Even if you don't merge this to master, making it a branch of github.com/mediocregopher/radix would allow people to reference that branch for at least *some* version of this repository.

It's understandably more ugly to have that /v3/ folder there, but it seems like the right approach while the Go community is in this transition period between dep and modules.

PR #137, opened 2019-07-07
Implement tracking of CLUSTERDOWN state 

### LINKED ISSUE #136: Handle CLUSTERDOWN and TRYAGAIN
When debugging #126 I noticed that there are 2 temporary cluster errors that radix currently does not handle:

* `CLUSTERDOWN`, which is returned when not all slots can be served and
* `TRYAGAIN` which is returned when trying multi key operations on a slot that is getting resharded

Currently radix returns these errors as is, but I propose we instead try to handle them similar to how we handle `ASK` and `MOVED` by just retrying the command (if it's retryable).

Also we should trace both of them via the new `trace` subpackage.

* For `TRYAGAIN` we can simply add a new field to `trace.ClusterRedirected`.
* For `CLUSTERDOWN`, which is a global state, I propose we add a new callback `Down` that receives the current topology as known to the `*Cluster` as a list of `trace.ClusterNodeInfo`.

nussjustin:
`CLUSTERDOWN` errors are now tracked in a `*Cluster` and changes to the state can be traced via a new callback in the `trace` subpackage. Also by default, when trying to execute commands on a cluster that is currently down the execution of the command will be delayed by a small, configurable time.

Also adds support for `CLUSTERDOWN` and `TRYAGAIN` in the stub cluster implementation. 

Fixes #136


cluster.go
```diff
@@ -135,6 +150,7 @@ func NewCluster(clusterAddrs []string, opts ...ClusterOpt) (*Cluster, error) {
 	defaultClusterOpts := []ClusterOpt{
 		ClusterPoolFunc(DefaultClientFunc),
 		ClusterSyncEvery(5 * time.Second),
+		ClusterWaitWhenDown(100 * time.Millisecond),
```
nussjustin: Random magic number incoming. This is just a random value that I came up with when prototyping, which worked great with the artificial delays I configured for my local cluster.

cluster.go
```diff
@@ -100,6 +112,8 @@ type Cluster struct {
 	// used to deduplicate calls to sync
 	syncDedupe *dedupe
 
+	down uint64
```
nussjustin: I used a atomic variable for two reasons:

1. It was quick to implement
2. Since this is potentially checked on every command invocation the check must not be slow especially under contention. As long as the value of the variable does not change all reads should be perfectly scalable.

I have no problem with replacing this by something else. Also I just realized that this is missing a big fat warning that this is accessed using the `sync/atomic` package...

cluster.go
```diff
@@ -459,13 +471,42 @@ func (c *Cluster) Do(a Action) error {
...
+			Moved:         moved,
+			Ask:           ask,
+			RedirectCount: doAttempts - attempts,
+		})
 	}
 }
 
 func (c *Cluster) doInner(a Action, addr, key string, ask bool, attempts int) error {
+	if c.co.clusterDownWait > 0 && c.isClusterDown() {
+		time.Sleep(c.co.clusterDownWait)
```
nussjustin: Again, simplest thing to implement for prototyping. In my test this works really great and avoids complex synchronization logic with other goroutines.

cluster_test.go
```diff
@@ -114,6 +121,12 @@ func TestClusterDo(t *T) {
 		cmd := Cmd(&vgot, "GET", k)
 		require.Nil(t, c.doInner(cmd, stub16k.addr, k, false, 2))
 		assert.Equal(t, v, vgot)
+		assert.Equal(t, trace.ClusterRedirected{
```
nussjustin: Leftovers from my `TRYAGAIN` changes, but since they don't hurt and help testing the tracing I decided too keep these.

cluster.go
```diff
@@ -82,6 +84,16 @@ func ClusterSyncEvery(d time.Duration) ClusterOpt {
 	}
 }
 
+// ClusterWaitWhenDown tells the Cluster to wait for the duration d before executing
+// commands when the cluster is down. If d is <= 0 the Cluster will not wait with
+// executing commands when the cluster is down.
+// Note that even if d is > 0, calls to Sync will not wait.
+func ClusterWaitWhenDown(d time.Duration) ClusterOpt {
```
nussjustin: Naming is hard

nussjustin:
No need to this to be marked as draft.

cluster.go
```diff
@@ -485,11 +526,19 @@ func (c *Cluster) doInner(a Action, addr, key string, ask bool, attempts int) er
 
 	err = p.Do(thisA)
 	if err == nil {
+		c.setClusterDown(false)
 		return nil
 	}
 
-	// if the error was a MOVED or ASK we can potentially retry
 	msg := err.Error()
```
mediocregopher: as long as the logic in this area is getting changed, might as well add:

```
if !errors.As(err, new(resp2.Error)) {
    return err
}
msg := err.Error()
```

I can't imagine what kind of network errors would start with strings like `CLUSTERDOWN` or `MOVED` or `ASK`, but might as well be safe.

nussjustin:
Will update the branch later today

> I don't really have a solution for this, except to have another ping routine on Cluster, which seems wrong, and I don't think it's worth holding up the PR for, but I figured I'd bring it up.

How about storing a unix timestamp of when last time we got a `CLUSTERDOWN`, with 0 meaning the cluster is up. In that case we could simply if the value is `>= 0` and `>= time.Now().Sub(<some time>).UnixTimestamp()` and only sleep if both conditions are true (the first one is just to avoid the cost of the time stuff).

This way on each new `CLUSTERDOWN` the timestamp would just get updated and once we succeed with a command we would just set the timestamp to 0.

cluster.go
nussjustin: Done

cluster.go
```diff
@@ -82,6 +84,16 @@ func ClusterSyncEvery(d time.Duration) ClusterOpt {
 	}
 }
 
+// ClusterWaitWhenDown tells the Cluster to wait for the duration d before executing
```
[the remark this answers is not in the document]
nussjustin: Done

nussjustin:
Ok, replaced the "boolean" with a `atomic.Value` holding a `time.Time` and and a `sync.Mutex` for writes. The code in `doInner` was updated to skip sleeping if there was already some time since the last `CLUSTERDOWN`.

cluster.go
```diff
@@ -459,13 +475,79 @@ func (c *Cluster) Do(a Action) error {
...
+	// avoid taking the lock when the down state was already cleared
+	prevVal, _ := c.lastClusterdown.Load().(time.Time)
+	if !down && prevVal.IsZero() {
+		return false
+	}
+
+	var newVal time.Time
+	if down {
+		// check for a small difference in time to reduce contention
+		if newVal = time.Now(); newVal.Sub(prevVal) < time.Millisecond {
```
nussjustin: If in the future lock contention becomes to noticeable we could increase the duration here (or even try to be smart with the configured sleep duration), but I think this should work well enough for now.

mediocregopher:
Instead of using an `atomic.Value`, couldn't we keep using an `int64` with `atomic` and store the unix timestamp in it? Then we don't need to worry about lock contention anymore.

nussjustin:
PTAL

I used a mutex since it was easy to get right and I, in my sleep deprived state, failed to write a version without mutex that I liked. But after a bit of pizza I now got a version that I like.

The new code avoids the mutex completely and uses a simple CAS loop with an extra load when the compare fails, to avoid the problem with the state going from down to up and down again.

cluster.go
```diff
@@ -459,13 +474,73 @@ func (c *Cluster) Do(a Action) error {
...
+		if prevVal == 0 || updatedVal == 0 || newVal == updatedVal {
+			return false
+		}
+		prevVal = updatedVal
+	}
+
+	changed = (prevVal == 0 && newVal != 0) || (prevVal != 0 && newVal == 0)
+
+	if changed && c.co.ct.StateChange != nil {
+		// we assume that this returns quick. we could also do this
```
mediocregopher: this comment might be out of date, since there's no lock anymore

cluster.go
```diff
@@ -459,13 +474,73 @@ func (c *Cluster) Do(a Action) error {
 	return c.doInner(a, addr, key, false, doAttempts)
 }
 
+func (c *Cluster) getClusterDownSince() int64 {
+	return atomic.LoadInt64(&c.lastClusterdown)
+}
+
+func (c *Cluster) setClusterDown(down bool) (changed bool) {
+	prevVal := atomic.LoadInt64(&c.lastClusterdown)
```
mediocregopher: So initially I was going to ask if we could just do an atomic swap here instead of a load and CAS, as the code would be a lot less complex, but then I did a benchmark and swap is ~10x less performant than load. Can you put a comment to that effect here, in case anyone comes along later wondering why this section is so much more complex than necessary?

cluster.go
```diff
@@ -459,13 +474,73 @@ func (c *Cluster) Do(a Action) error {
...
+		}
+	}
+
+	for !atomic.CompareAndSwapInt64(&c.lastClusterdown, prevVal, newVal) {
+		updatedVal := atomic.LoadInt64(&c.lastClusterdown)
+		// if another goroutine changed the state before we could, we can just return
+		// and avoid more CAS operations since the important thing (the state change)
+		// already happened and the difference in newVal and updatedVal will be small
+		// (when the cluster is down)
+		if prevVal == 0 || updatedVal == 0 || newVal == updatedVal {
```
mediocregopher: I'm having trouble following this part. Why are we able to assume there wasn't a change if `prevVal` is 0? On its own that doesn't seem like it would have any bearing on whether or not a change had happened.

nussjustin: Added a comment that explains this, but I realized that this isn't really necessary since we can just use a swap which is much simpler and easier to get right and has basically the same performance. But yeah the commit with the comment is still there.

cluster.go
nussjustin: After a bit of sleep, reading your comments and running a few benchmarks, I realized that we can remove the CAS-loop and replace it with a swap without losing on performance, but winning on simplicity and readability.

Also added two comments to explain the logic a bit better.

cluster.go
nussjustin: Done

mediocregopher:
Nice, I guess we'll have to watch out for issues about that race condition, if people complain about it we might have to go back to the more complex handling. But this seems fine for now, definitely better than not having it.

Issue #131, opened 2019-06-25
Update documentation


mediocregopher:
Specifically regarding #129 and the dichotemy between `CmdAction`'s `Run` and `MarshalRESP`/`UnmarshalRESP`. Someone implementing a new `CmdAction` needs to understand exactly how those two are used so that they can navigate the pipeline infested waters of the Pool implementation.

Also the rest of the docs could probably use a polish, off the top of my head there's nothing in the docs that actually explains what implicit pipelining _is_.

mediocregopher:
@cep21 I updated the docs in question, can you let me know if they make more sense to you and might have helped you with the problems you were having in #129 ?

PR #138, opened 2019-07-07
Document implicit pipelining

nussjustin:
Updates #131

### DIFF
--- a/radix.go
+++ b/radix.go
@@ -142,6 +142,29 @@
 //
 // Use the golang.org/x/xerrors package if you're using an older version of go.
 //
+// Implicit pipelining
+//
+// Implicit pipelining is an optimization that reduces the overhead of
+// I/O and the CPU usage for marshalling and unmarshalling on both the
+// client and server when executing many concurrent commands on a Client.
+//
+// Commands that can be safely pipelined (currently all commands except those
+// that may block) will be delayed by a small amount of time and be send to
+// redis in a single batch, similar to manually using a Pipeline. All other
+// commands will be executed without delay as if implicit pipelining was
+// disabled.
+//
+// Currently only pools returned by NewPool implement implicit pipelining.
+// This includes the pools used by default in NewCluster and NewSentinel,
+// thus both Cluster and Sentinel can profit from implicit pipelining.
+//
+// See the documentation on Pool for more information about the current
+// implementation of implicit pipelining and for how to configure or disable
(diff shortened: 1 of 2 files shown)


radix.go
```diff
@@ -142,6 +142,29 @@
...
+//
+// Currently only pools returned by NewPool implement implicit pipelining.
+// This includes the pools used by default in NewCluster and NewSentinel,
+// thus both Cluster and Sentinel can profit from implicit pipelining.
+//
+// See the documentation on Pool for more information about the current
+// implementation of implicit pipelining and for how to configure or disable
+// the feature.
+//
+// For a performance comparisons between Clients with and without impliciz
```
cep21: Spelling `impliciz`

radix.go
```diff
@@ -142,6 +142,29 @@
 //
 // Use the golang.org/x/xerrors package if you're using an older version of go.
 //
+// Implicit pipelining
+//
+// Implicit pipelining is an optimization that reduces the overhead of
```
mediocregopher: I try to keep the main package doc very high level, and stick to only the bare minimum info that someone needs in order to use radix, or to preemptively answer questions that we get a lot (e.g. about AUTH). So I think these three paragraphs are a bit too wordy, but could be rolled together into one, like this:

> Implicit pipelining is an optimization implemented and enabled in the default Pool implementation (and therefore also used by Cluster and Sentinel) which involves delaying concurrent Cmds and FlatCmds a small amount of time and sending them to redis in a single batch, similar to manually using a Pipeline. By doing this radix significantly reduces the I/O and CPU overhead for concurrent requests.

> Note that only commands which do not block are eligible for implicit pipelining.

> See the documentation on Pool etc etc

What do you think?

mediocregopher:
@cep21 did you have a chance to check out the changes I made to the `CmdAction` docs? I tried to make it clearer how `CmdAction` interacts with pipelining (implicit or otherwise), hopefully I did a better job than before.

radix.go
nussjustin: Done

radix.go
nussjustin: Sounds good to me. I updated the branch.

cep21:
@mediocregopher I'm looking it over.  Give me a few days and I'll get back to you with my thoughts.

cep21:
Oops dropped the ball here :) Thanks for the comments!

Issue #126, opened 2019-06-17
Investigating error when unmarshal failed


kimtree:
Hi,
I just migrate our Redis library to latest Radix commit. We had an issue today while the cluster is changing. I'm sorry I couldn't get detailed logs since changing cluster is the other team's role. Below is the code where the error occurred.

```golang
var result []int
err := cluster.Do(radix.Cmd(&result, "HKEYS", key))
if err != nil {
	return nil, err
}
return result, err
```

The error messages we could get were three kinds. "can't unmarshal into *[]int", "cannot decode redis array into int", "unknown type prefix '0'". This occurs while cluster failover and I'm not sure why this kind of errors occurs. This issue resolved after we restart our server to re-connect Redis server. It might be better if we could know the exact result from the Redis for investigating this issue.

This also might be a clue. (error from redis)
```
-MASTERDOWN Link with MASTER is down and slave-serve-stale-data is set to 'no'.
```

nussjustin:
> This occurs while cluster failover and I'm not sure why this kind of errors occurs.

Was this a manual failover or did a master node die?

The errors all look like you either got some garbage/partial response from Redis or some other code failed and didn't fully read the response from Redis.

I'm not sure how the connection would end up in this state, since the Pool connections (`*ioErrConn`) should detect network errors (broken connections) and then remove the connections from the pool. The automatic pipelining came to my mind for a second, but since the pipeliner also uses the `*ioErrConn` networks errors should still be handled correctly.

Still in either of of these cases a command (and possibly even all commands after that) would fail when reading because there is still a unread partial response in the buffer (either in the `*bufio.Reader` or the socket buffer).

Since most responses are short there is a high chance that the data left in the buffer begins with a type prefix (e.g. `$`) which would lead radix to assume a specific type. This could explain the different error messages.

@mediocregopher Any ideas? Or do you see any problem with what I wrote?

kimtree:
@nussjustin Yes, I assume this was a manual failover and maybe it made a master node to be shown dead. This failover progress was done by the other team, so I'm not sure what's going on inside. 

We didn't use automatic pipelining at the moment. These errors still occur intermittently even though cluster failover progress already finished. I have no idea why this garbage connections still exists... :(

mediocregopher:
[...]

PR #139, opened 2019-07-08
Handle I/O errors when using implicit pipelining

nussjustin:
Implicit pipelining didn't account for I/O errors (both connection-level errors and other errors that leave the connection in a state were we can't safely continue using the connection).

This fixes the error handling by reusing the logic from `ioErrConn` and returning stopping the unmarshaling of later commands once the connection ends up in an unrecoverable state.

See [my comment on 126](https://github.com/mediocregopher/radix/issues/126#issuecomment-509144660) and the comments above that for more information about the problem this solves.

While I was here I added a check to `(*ioErrConn.Read)` that avoids the somewhat expensive `errors.As` call when there is no error.

Updates #126


mediocregopher:
Re: the two options you gave in #126, I think this is definitely the better one. Option 1 involves waiting on a connection which has already shown itself to be faulty, might as well bail.

The code looks good, not sure why one of the tests isn't passing (it's getting the error `XADD command not found` or something, which is bizarre....), but once it does I'll merge.

nussjustin:
```
$ sudo apt-get install -y redis-server
Reading package lists... Done
Building dependency tree       
Reading state information... Done
redis-server is already the newest version (4:4.0.6-1chl1~trusty1).
0 upgraded, 0 newly installed, 0 to remove and 296 not upgraded.
```

Looks like we don't get Redis 5 sometimes? Rerunning the tests would probably fix this, but this is still strange.

mediocregopher:
Damnit travis. I reran it and that did fix it. Gonna look over it one more time and then merge.

PR #142, opened 2019-07-11
Move the traceRedirected call in Cluster to be just before the redire…

mediocregopher:
…ct actually happening.

Previously there were a few cases where traceRedirected would be called but the redirect wouldn't actually occur. This fixes that.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -540,14 +540,15 @@ func (c *Cluster) setClusterDown(down bool) (changed bool) {
 	return changed
 }
 
-func (c *Cluster) traceRedirected(addr, key string, moved, ask bool, attempts int) {
+func (c *Cluster) traceRedirected(addr, key string, moved, ask bool, count int, final bool) {
 	if c.co.ct.Redirected != nil {
 		c.co.ct.Redirected(trace.ClusterRedirected{
 			Addr:          addr,
 			Key:           key,
 			Moved:         moved,
 			Ask:           ask,
-			RedirectCount: doAttempts - attempts,
+			RedirectCount: count,
+			Final:         final,
 		})
 	}
 }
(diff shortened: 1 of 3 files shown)


nussjustin:
LGTM in general, but I was thinking if you considered also tracing when we can not retry (basically when not `ccra.ClusterCanRetry()`).

In that case the user now get's the original error, so they could track it manually, but it would probably be easier for most users if these cases were also traced (maybe with an extra flag).

WDYT?

mediocregopher:
> In that case the user now get's the original error, so they could track it manually, but it would probably be easier for most users if these cases were also traced (maybe with an extra flag).

I think it's worth drawing the line for tracing at "events and information users couldn't get themselves". In this case the error would be returned to the user and they'd know it couldn't have been retried because their type didn't implement it.

Is there a particular use-case you have in mind where having it be a trace would make it easier to deal with for the user?

nussjustin:
> I think it's worth drawing the line for tracing at "events and information users couldn't get themselves". In this case the error would be returned to the user and they'd know it couldn't have been retried because their type didn't implement it.

If that's were you want to draw the line, I'm okay with that.

I just imagined what would happen if users trace the redirects (e.g. for monitoring), but don't also manually trace the error returned for the last (non-traced) redirect (which would involve checking the error as `doInner` does). In that case it could look like the cluster state was always corrected before the last attempt and that the last attempt succeeded.

mediocregopher:
Ok that's a good point, since we don't expose `doAttempts` it's difficult to know when the redirecting has ended. I added a `Final` flag to `trace.ClusterRedirect`, and moved the trace slightly so it happens before the `attempts` check. I also realized the math on `RedirectCount` wasn't quite right, so I fixed that.

nussjustin:
LGTM

Issue #149, opened 2019-07-22
EvalAction never runs EVAL, only EVALSHA


everettcaleb:
When I run my EvalAction.Cmd(), it always fails with the following "NOSCRIPT" error:

```
NOSCRIPT No matching script. Please use EVAL.
```

Here's a snippet that shows how I'm using it (I cannot show the real code since this is for work):

```go
var data string
if err := pool.Do(script.Cmd(&data, "MYKEY", strconv.FormatInt(time.Now().Unix(), 10))); err != nil {
	return nil, fmt.Errorf("failed to run Lua script: %v", err)
}
if data == "" {
	return nil, nil
}
```

Where `pool` is a `*radix.Pool` and `script` is a `radix.EvalScript`.

While troubleshooting, I modified some code in `action.go` locally to get more debug info. Here's the code with my changes (just `MarshalRESP` and `Run`):

```go
func (ec *evalAction) MarshalRESP(w io.Writer) error {
	fmt.Printf("evalAction.MarshalRESP: %+v\n", ec) // ADDED

	// EVAL(SHA) script/sum numkeys args...
	if err := (resp2.ArrayHeader{N: 3 + len(ec.args)}).MarshalRESP(w); err != nil {
		return err
	}

	var err error
	if ec.eval {
		fmt.Println("USING EVAL") // ADDED
		err = marshalBulkStringBytes(err, w, eval)
		err = marshalBulkString(err, w, ec.script)
	} else {
		fmt.Println("NOT USING EVAL") // ADDED
		err = marshalBulkStringBytes(err, w, evalsha)
		err = marshalBulkString(err, w, ec.sum)
	}

	err = marshalBulkString(err, w, strconv.Itoa(ec.numKeys))
	for i := range ec.args {
		err = marshalBulkString(err, w, ec.args[i])
	}
	return err
}

func (ec *evalAction) Run(conn Conn) error {
	run := func(eval bool) error {
		ec.eval = eval
		fmt.Println("evalAction func run eval=", eval) // ADDED
		fmt.Printf("evalAction: %+v\n", ec) // ADDED
		if err := conn.Encode(ec); err != nil {
			return err
		}
		return conn.Decode(resp2.Any{I: ec.rcv})
	}

	err := run(false)
	if err != nil && strings.HasPrefix(err.Error(), "NOSCRIPT") {
		err = run(true)
	}
	return err
}
```

This is the (script redacted) log output I get when I run the test code snippet:

```
evalAction func run eval= false
[...]

PR #151, opened 2019-07-22
Patch to handle NOSCRIPT errors from #149 

everettcaleb:


everettcaleb:
Perhaps this approach for #149 ?

PR #161, opened 2019-08-21
add an Action field to trace.PoolDoCompleted

### LINKED ISSUE #159: trace.PoolDoCompleted could include the Action
This would open up a lot of flexibility in allowing users to trace out specific parts of their codebase.

mediocregopher:
Fixes #159 

@fastest963 here's this. The `Action` field has to be an `inteferface{}` since the `trace` package can't import the main `radix` package (due to import loops). Since there's not much someone could do with the `Action` without doing a type-cast on it anyway I don't think this matters much.

### DIFF
--- a/pool.go
+++ b/pool.go
@@ -602,7 +602,7 @@ func (p *Pool) Do(a Action) error {
 	startTime := time.Now()
 	if p.pipeliner != nil && p.pipeliner.CanDo(a) {
 		err := p.pipeliner.Do(a)
-		p.traceDoCompleted(time.Since(startTime), err)
+		p.traceDoCompleted(a, time.Since(startTime), err)
 
 		return err
 	}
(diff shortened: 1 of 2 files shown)


nussjustin:
Haven't checked yet, but doesn't this make the assumption that you can reuse an `Action`?

In most cases the `Action` will be of type `*cmdAction` and these must not be accessed after calling `UnmarshalRESP` since they are put back into a `sync.Pool`.

fastest963:
We're still going to need to wrap all actions because the default `*cmdAction` we can't actually do anything with since its private. If you could add a `Command() string` method to it then we could at least cast it as an interface with a Command method and call that.

mediocregopher:
> Haven't checked yet, but doesn't this make the assumption that you can reuse an Action?

Oh yeah, that's a good point.

> We're still going to need to wrap all actions

Right, that was part of my assumption. I don't think we're going to be able to do anything without users wrapping commands. `*cmdAction` implements `fmt.Stringer`, so at least you'd be able to use that?

> If you could add a Command() string method to it then we could at least cast it as an interface with a Command method and call that.

I'm not sure I like `Command()` as a method, it's too specific, but that gives me another idea: have `PoolDoCompleted` have a field like `ActionString` instead, which is a string representation of the action as returned by `fmt.Sprint`. By default this will work with `*cmdAction` just fine, and if someone wants to wrap the `Action` to include more information they could do that too.

The only hitch is we'd have to change how `cmdAction`'s `String` method works, because right now I think it's highly inefficient.

What do you all think? It's still a bit janky, but it could work maybe?

fastest963:
> I'm not sure I like Command() as a method, it's too specific, but that gives me another idea: have PoolDoCompleted have a field like ActionString instead, which is a string representation of the action as returned by fmt.Sprint. By default this will work with *cmdAction just fine, and if someone wants to wrap the Action to include more information they could do that too.

That's not bad, but I'd still need to split on spaces or something and get out the actual command. What's nice about gocql's [ObservedQuery](https://godoc.org/github.com/gocql/gocql#ObservedQuery) is that it has the Statement (and keyspace) which are perfect for metrics. I don't know what the use case would be for the whole Action string other than parsing it to find the Command itself. If that's all people are doing I'd rather have the normal Action and just call String myself but at least I'd be able to wrap it with my own type and cast it if need be (versus parsing a string).

PR #168, opened 2019-11-06
Defer connection closure outside critical section

### LINKED ISSUE #166: Cluster sync may takes times when a node is stopped
Hello,
I noticed doing chaos testing that request may be block on the addrForKey, because the sync was taking time (around 10ms) to close the connection to a node that is stopping.

I will submit a PR to defer the close after the swap of the pool so that the lock can be released.

dprofeta:
When there is a redis node that is stopped/crash, the closure of the
connection can take longer than usual. As the lock on the connection
was taken, the client cannot send any message to the cluster during
this time.

While doing chaos testing, I observed hundreds of milliseconds delay
to send the next message. To prevent that, the closure of the
connection is defered to after the critical section.

This close #166

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -378,19 +378,24 @@ func (c *Cluster) sync(p Client) error {
 
 	c.traceTopoChanged(c.topo, tt)
 
-	// this is a big bit of code to totally lockdown the cluster for, but at the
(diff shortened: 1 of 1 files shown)


mediocregopher:
LGTM, thanks!

PR #171, opened 2019-11-12
Fix alignment for atomic values in structs

### LINKED ISSUE #169: unable to create connection on Windows 32 bit machine's with GO MODULES
I am unable to create the connection to the running redis when the go application is designed as GO MODULES on windows 32 bit computer. it works fine when being compiled and executed on 64 bit computer.

on 32 bit computer's golang module it's connect and immediately disconnect from redis.

nussjustin:
Checked via github.com/dominikh/go-tools/cmd/structlayout

Fixes #169


mediocregopher:
Thanks for looking into this @nussjustin  :)

PR #173, opened 2019-11-19
Fix closing of Sentinel instances while updating state

### LINKED ISSUE #172: Locking on sentinel close
Hi,

I code redis client that use sentinel but it seems lock on wait group while calling Close()

	package main

	import (
		"github.com/mediocregopher/radix"
		"log"
	)

	func main() {
		sentinel, err := radix.NewSentinel("mymaster", []string{"127.0.0.1:26379", "127.0.0.2:26379"})

		if err != nil {
			log.Println(err)
			return
		}

		defer sentinel.Close()

		if err := sentinel.Do(radix.Cmd(nil, "PUBLISH", "pub", "this is a message")); err != nil {
			log.Println(err)
		}

		log.Println("finish")
	}

The code will not exit after print finish.

nussjustin:
Fixes #172

### DIFF
--- a/sentinel.go
+++ b/sentinel.go
@@ -3,6 +3,7 @@ package radix
 import (
 	"net"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	errors "golang.org/x/xerrors"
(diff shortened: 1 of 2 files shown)


sentinel.go
```diff
@@ -468,9 +473,17 @@ func (sc *Sentinel) innerSpin() error {
 			// loop
 		case <-sc.pconnCh:
 			switchMaster = true
+			if waitFor := atomic.SwapUint32(&sc.testSleepBeforeSwitch, 0); waitFor > 0 {
```
nussjustin: @mediocregopher this is kinda hacky but I couldn't think about a better way for this

mediocregopher:
Thanks @nussjustin ! LGTM but it's not building on 1.12, I guess `Milliseconds` is a new thing in 1.13.

nussjustin:
@mediocregopher Fixed, PTAL

PR #180, opened 2019-12-20
Drain pipeline on decode error

### LINKED ISSUE #179: Pipeline decode error results in unusable connection
Whenever a decoding error happens as part of a `radix.Pipeline` command the connection is left an unusable state. Subsequent commands return results for commands that were part of the pipeline. The code below illustrates the problem. 

```go
package main

import (
	"fmt"
	"strings"

	"github.com/mediocregopher/radix/v3"
)

func populateDb(conn radix.Conn, kvs map[string]string) error {
	// Execute as a single pipeline
	var cmds []radix.CmdAction

	for k, v := range kvs {
		cmds = append(cmds, radix.Cmd(nil, "SET", k, v))
	}

	return conn.Do(radix.Pipeline(cmds...))
}

func main() {
	conn, err := radix.Dial("tcp", "localhost:6379")
	if err != nil {
		panic(err)
	}

	kvs := map[string]string{
		"foo": "str",
		"bar": "another str",
	}

	if err := populateDb(conn, kvs); err != nil {
		panic(err)
	}

	var intRcv int
	var strRcv string

	pipeline := radix.Pipeline(
		radix.Cmd(&intRcv, "GET", "foo"),
		radix.Cmd(&strRcv, "GET", "bar"),
	)

	if err := conn.Do(pipeline); err == nil || !strings.HasPrefix(err.Error(), "failed to decode pipeline") {
		panic(err)
	}

	if err := conn.Do(radix.Cmd(&strRcv, "GET", "foo")); err != nil {
		panic(err)
	}

	fmt.Printf("Expected: %s\n", kvs["foo"])
	fmt.Printf("Got: %s\n", strRcv)

}
```

Prints:
```
Expected: str
Got: another str
```

maciej:
Resolves #179

### DIFF
--- a/action.go
+++ b/action.go
@@ -441,12 +441,21 @@ func (p pipeline) Run(c Conn) error {
 	if err := c.Encode(p); err != nil {
 		return err
 	}
-	for _, cmd := range p {
+
+	for i, cmd := range p {
 		if err := c.Decode(cmd); err != nil {
+			p.drain(c, len(p)-i-1)
 			return decodeErr(cmd, err)
 		}
 	}
 	return nil
+}
+
+func (p pipeline) drain(c Conn, n int) {
+	rcv := resp2.Any{I: nil}
+	for i := 0; i < n; i++ {
+		_ = c.Decode(&rcv)
+	}
 }
 
 func decodeErr(cmd CmdAction, err error) error {
--- a/action_test.go
+++ b/action_test.go
@@ -6,9 +6,10 @@ import (
 	"fmt"
 	. "testing"
 
-	"github.com/mediocregopher/radix/v3/resp/resp2"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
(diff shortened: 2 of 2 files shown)


action.go
```diff
@@ -441,14 +441,23 @@ func (p pipeline) Run(c Conn) error {
...
 		if err := c.Decode(cmd); err != nil {
+			p.drain(c, len(p)-i-1)
 			return decodeErr(cmd, err)
 		}
 	}
 	return nil
 }
 
+func (p pipeline) drain(c Conn, n int) {
+	var rcv resp2.RawMessage
```
mediocregopher: Rather than a `resp2.RawMessage`, using a `resp2.Any{I: nil}` will discard the next message on the wire, rather than reading it into a buffer and incurring an allocation.

maciej: Fixed! I'm amended the existing commit.

mediocregopher:
Merged, thanks again :)

Issue #189, opened 2020-02-21
FlatCmd causes panic


ollevche:
Hi! I don't sure if it's undefined behavior or i misused the API, but here is my problem. I have a struct with a field of pointer type and this field causes panic in FlatCmd. Is it a bug?

Here is the code illustrating the problem:

```
package main

import (
	"fmt"
	"github.com/mediocregopher/radix/v3"
)

func main() {
	p, err := radix.NewPool("tcp", ":6379", 2)
	if err != nil {
		fmt.Println(err)
		return
	}

	obj := struct {
		// StrPtrField *string // uncomment this line for panic
		IntField    int
	}{
		IntField: 1,
	}

	err = p.Do(radix.FlatCmd(nil, "HMSET", "key", obj))
	if err != nil {
		fmt.Println(err)
		return
	}
}
```

mediocregopher:
Hi @ollevche, that code definitely should not cause a panic, and it definitely does on my machine as well. Good find, and thanks for the easy reproduction steps :) I will try to track it down in the next week.

ollevche:
There's one more thing to mention. It's impossible to recover the panic from client side code. I guess, your code doesn't `recover()` goroutines it started, so any undiscovered bug with panic is able to shut down an entire app. Example code:

```
package main

import (
	"fmt"
	"github.com/mediocregopher/radix/v3"
)

func main() {
	p, err := radix.NewPool("tcp", ":6379", 2)
	if err != nil {
		fmt.Println(err)
		return
	}

	obj := struct {
		StrPtrField *string // uncomment this line for panic
		IntField int
	}{
		IntField: 1,
	}

	defer func() {
		if r := recover(); r != nil {
			fmt.Println("recovered panic")
		}
	}()

	err = p.Do(radix.FlatCmd(nil, "HMSET", "key", obj))
	if err != nil {
		fmt.Println(err)
		return
	}
}

```

What's your point on this? Placing `recover()` calls everywhere?

PR #190, opened 2020-02-26
Simplify and improve pipeliner error handling

nussjustin:
As reported in #189, the current `pipeliner` implementation does not handle panics when marshalling/unmarshalling and will kill the whole program. Fix this by recovering any panics from within the `pipelinerPipeline.Run` method.

Since I am here already I also refactored the error handling for the pipeliner which was more complex and subtle than it should have been.

Updates #189

### DIFF
--- a/pipeliner.go
+++ b/pipeliner.go
@@ -1,10 +1,13 @@
 package radix
 
 import (
+	"bufio"
 	"fmt"
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/mediocregopher/radix/v3/resp"
 )
 
 var blockingCmds = map[string]bool{
@@ -176,14 +179,11 @@ func (p *pipeliner) flush(reqs []CmdAction) []CmdAction {
 			p.reqsBufCh <- reqs[:0]
 		}()
 
-		err := p.c.Do(pipelinerPipeline{
-			pipeline: pipeline(reqs),
-		})
+		pp := &pipelinerPipeline{pipeline: pipeline(reqs)}
+		defer pp.flush()
 
-		for _, req := range reqs {
-			cmd := req.(*pipelinerCmd)
-			cmd.setRes(err)
-			cmd.send()
+		if err := p.c.Do(pp); err != nil {
+			pp.doErr = err
 		}
 	}()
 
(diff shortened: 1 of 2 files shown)


pipeliner.go
```diff
@@ -194,15 +192,32 @@ func (p *pipeliner) flush(reqs []CmdAction) []CmdAction {
...
+
+func (p *pipelinerCmd) setRes(err error) {
+	if p.resSet {
+		return
+	}
+	p.res = err
+	p.resSet = true
+}
+
+func (p *pipelinerCmd) send() {
```
mediocregopher: Just for clarity I think `sendRes` might be a better name. Also, are there any cases where `resSet == true` and `p.res == nil`? I can't think how that would happen with the given logic, but maybe I'm missing something. If not, maybe it'd be simpler to drop `resSet` and just check that `res == nil` before writing it.

nussjustin: `p.res` will be `nil` whenever the was no error, which should be most cases. If we were to drop `resSet` we would need to keep track where the `pipelinerPipine` exits with an error, otherwise we may set an error on a `pipelinerCmd` that had no error before.

Maybe another name than `resSet` might be better.

nussjustin:
I think we should do both. Handling panics in `resp.Any` is more generally useful, but having it also in the `pipeliner` does not hurt (it's basically just a `defer` with an `if`).

nussjustin:
Updated the PR with a (IMO) cleaner error tracking.

I also played with using values instead of pointers for `pipelinerCmd`s and found this to be a) a bit easier to read and b) much simpler since this automtically avoids all race conditions with the `*pipelinerCmd`s since it only pools the channels. I pushed this to [another branch here](https://github.com/nussjustin/radix/commit/8372478da811aec5c7c7e87b132d35eb0cd4c07a). The commit is a bit large because I renamed a few things. The only negative I could see about this is that I couldn't use the existing `pipeline` for this anymore and instead hat to re implement it (2 methods).

mediocregopher:
Reading through both I kind of like the current way a bit better, mostly because it's not re-implementing `Keys()`, and what you have here is easy enough to follow imo. But I'm down to merge whichever you prefer, both lgtm :)

nussjustin:
Let's go with the branch from this PR then for now. The change is smaller and simpler.

mediocregopher:
Done!

PR #191, opened 2020-02-26
Wrap detailed pipeline errors

nussjustin:
Using the `%w` verb the original errors can be checked/extracted using `Is/As`.


mediocregopher:
Thanks! Can't wait to be able to drop xerrors...

PR #192, opened 2020-02-26
Fix marshalling of nil pointers in structs

### LINKED ISSUE #189: FlatCmd causes panic

nussjustin:
Fixes #189

### DIFF
--- a/resp/resp2/resp.go
+++ b/resp/resp2/resp.go
@@ -674,7 +674,13 @@ func (a Any) MarshalRESP(w io.Writer) error {
 
 	// if it's a pointer we de-reference and try the pointed to value directly
 	if vv.Kind() == reflect.Ptr {
-		return a.cp(reflect.Indirect(vv).Interface()).MarshalRESP(w)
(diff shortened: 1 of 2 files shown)


mediocregopher:
LGTM, thanks for jumping on this!

PR #195, opened 2020-03-14
Add support for usernames when authenticating

### LINKED ISSUE #188: Support authentication with username and password
Redis 6 will introduce the concept of different users as part of the new ACL support.

We currently only support configuring a password via `DialAuthPass` and ignore usernames set via the URL/DSN.

I propose we...

1. Add a new function `DialAuthUser` that takes a username and a password.
2. Stop ignoring the username provided via URL/DSN.
3. Allow setting the username via URL parameter

If both a username and password is provided (either via `DialAuthUser` or via DSN/URL) radix should automatically try sending both. If only a password and no username is provided it should just send the password. This should work since Redis 6 still supports the old `AUTH` syntax (the old syntax `AUTH <password>` gets interpreted as `AUTH default <password>`).

This will break users with Redis versions < 6 that set a username other than "default" in their URL/DSN, but this shouldn't be a problem and I assume most people don't do this.

nussjustin:
Fixes #188

### DIFF
--- a/radix.go
+++ b/radix.go
@@ -335,6 +335,8 @@ func DialTimeout(d time.Duration) DialOpt {
 	}
 }
 
+const defaultAuthUser = "default"
+
 // DialAuthPass will cause Dial to perform an AUTH command once the connection
 // is created, using the given pass.
 //
@@ -344,7 +346,7 @@ func DialTimeout(d time.Duration) DialOpt {
 // Using DialAuthPass is equivalent to calling DialAuthUser with user "default"
 // and is kept for compatibility with older package versions.
 func DialAuthPass(pass string) DialOpt {
-	return DialAuthUser("default", pass)
+	return DialAuthUser(defaultAuthUser, pass)
 }
 
 // DialAuthUser will cause Dial to perform an AUTH command once the connection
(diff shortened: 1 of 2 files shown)


radix.go
```diff
@@ -404,7 +417,18 @@ func parseRedisURL(urlStr string) (string, []DialOpt) {
 
 	var opts []DialOpt
 	q := u.Query()
-	if p, ok := u.User.Password(); ok {
+
+	n := u.User.Username()
```
mediocregopher: This whole section could probably be cleaned up into something like:

```go
const defaultAuthUser = "default" // also used in DialAuthPass

// ...

username := defaultAuthUser
if n := u.User.UserName(); n != "" {
    username = n
} else if n := q.Get("username"); n != "" {
    username = n
}

password := q.Get("password")
if p := u.User.Password(); p != nil {
    password = p
}

if password != "" {
    opts = append(opts, DialAuthUser(username, password))
}
```

nussjustin: PTAL

I removed the condition around the `DialAuthUser` call since `Dial` already handles the default user and empty passwords

mediocregopher:
LGTM, just needs the tests to be fixed =/ not sure if there's really a nice way to do it, might be necessary to use a Conn stub

nussjustin:
Ah yeah, didn't think about the syntax/argument error on Redis < 6... Will take a look at using a stub later this week

nussjustin:
Well I changed the tests to only check the username when Redis 6 is detected, since using a Stub doesn't really work here (`Dial` always connects to an actual socket).

The tests should probably be refactored at some point in the future to actually test what is send to redis, but I can't find the motivation to do this right now.

mediocregopher:
Lol, fair enough. Thanks!

PR #196, opened 2020-03-14
Avoid allocating a []byte when reading stream names

nussjustin:
Removes one allocation when calling StreamReader.Next:

```
name            old alloc/op   new alloc/op   delta
StreamReader-8    14.6kB ± 0%    14.5kB ± 0%  -0.22%  (p=0.008 n=5+5)

name            old allocs/op  new allocs/op  delta
StreamReader-8       146 ± 0%       145 ± 0%  -0.68%  (p=0.008 n=5+5)
```

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -355,7 +355,7 @@ func (sr *streamReader) Next() (stream string, entries []StreamEntry, ok bool) {
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


mediocregopher:
Nice! good find

Issue #174, opened 2019-11-20
Implement random replica reads for Sentinel and Cluster


nussjustin:
Currently commands on Sentinel and Cluster instances are always send to the corresponding master node. For users with high traffic, especially when including many writes, the master can quickly become a bottleneck. A common way to work around this is sending (some) reads to replicas instead. This lifts some load from the master while increasing the chance for stale data (since Cluster and Sentinel are both AP systems the possibility is already there even when reading from the master).

Doing this with the current API is possible but also quite cumbersome. I propose that we add a way for both `Cluster` and `Sentinel` to make replica reads on random replicas, possibly through an extra method `Read` that works like `Do` but sends the command to a replica.

For Sentinel the nodes must be configured with `replica-read-only` to allow reads from replicas which could just be documented.

For Cluster each connection must manually call `READONLY` once, which is hard to implement with the current API and may be something we would need to require from the users to do themselves with a customer ClientFunc.

mediocregopher:
(Sorry it took so long to respond to this -_-)

Starting with Sentinel seems to be the easiest option. I would suggest the name `DoSecondary` rather than `Read`; it's a bit wordier but it's more accurate, and conveys that the method is related to the `Do` method.

The `READONLY` thing is hard. Even if the user does use a custom `ClientFunc` there's the issue that the current Cluster implementation will promote a secondary Pool to a primary one during a failover event, if it can, rather than tear them down and start from scratch. But if all the Conns in the secondary Pool are in read-only mode, however they got in that mode, Cluster needs to be aware of that so it knows not to use the Pool.

nussjustin:
> But if all the Conns in the secondary Pool are in read-only mode, however they got in that mode, Cluster needs to be aware of that so it knows not to use the Pool.

That shouldn't be necessary.

Contrary to its name the `READONLY` command does not make the connection read only (yeah, I know...). If you send `READONLY` on a connection you can still write to it, assuming it's a master. This also works when the server was still a replica when the connection was put into `READONLY` mode and only became a master after that, while the connection is still open.

You can take a look at the code that handles the `READONLY` flags for clusters here: https://github.com/antirez/redis/blob/cf6cbbb881556d0665e3c1a417ee1eacb0be857f/src/cluster.c#L5667-L5677

mediocregopher:
Crazy >_< I wonder if there's any harm in just setting READONLY on all cluster connections then....

nussjustin:
[...]

PR #197, opened 2020-03-15
Add support for secondary reads with sentinel

nussjustin:
First part of #174 

Updates #174

### DIFF
--- a/sentinel.go
+++ b/sentinel.go
@@ -205,6 +205,31 @@ func (sc *Sentinel) Do(a Action) error {
 	return sc.clients[sc.primAddr].Do(a)
 }
 
+// DoSecondary is like Do but executes the Action on a random replica if possible.
+//
+// For DoSecondary to work, replicas must be configured with replica-read-only
+// enabled, otherwise calls to DoSecondary may by rejected by the replica.
+//
+// NOTE it's possible that in between Do being called and the Action being
+// actually carried out that there could be a failover event. In that case, the
+// Action will likely fail and return an error.
+func (sc *Sentinel) DoSecondary(a Action) error {
+	sc.l.RLock()
+	var addr string
+	for addr = range sc.clients {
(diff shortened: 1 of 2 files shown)


sentinel.go
```diff
@@ -205,6 +205,31 @@ func (sc *Sentinel) Do(a Action) error {
 	return sc.clients[sc.primAddr].Do(a)
 }
 
+// DoSecondary is like Do but executes the Action on a random replica if possible.
+//
+// For DoSecondary to work, replicas must be configured with replica-read-only
+// enabled, otherwise calls to DoSecondary may by rejected by the replica.
+//
+// NOTE it's possible that in between Do being called and the Action being
```
mediocregopher: in between `DoSecondary` being called *

sentinel.go
```diff
@@ -205,6 +205,31 @@ func (sc *Sentinel) Do(a Action) error {
...
+	sc.l.RLock()
+	var addr string
+	for addr = range sc.clients {
+		if addr != sc.primAddr {
+			break
+		}
+	}
+	sc.l.RUnlock()
+
+	c, err := sc.Client(addr)
```
mediocregopher: Right now this `DoSecondary` method is just a convenience helper, it doesn't do anything that the user couldn't do themselves. _But_ if you modified how `Client` works a bit you could remove an `RLock()`/`RUnlock()` in the path. Basically, move most of `Client`'s code into a new method, `clientInner` or something like that, and have that new method also accept an empty address string which it interprets to mean "random secondary". Then both `Client` and `DoSecondary` can call that directly. It might be worthwhile to have the new `Client` method check for being given an empty address? Not 100% on that.

nussjustin: Done

sentinel.go
nussjustin: Done

mediocregopher:
Nice, thanks!

PR #198, opened 2020-03-15
Add support for secondary reads with cluster

nussjustin:
Second part of #174

This currently requires users to manually send the `READONLY` command using a custom `ClusterPoolFunc` since we can not automatically send `READONLY` to all connections.

Updates #174

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -123,6 +123,7 @@ type Cluster struct {
 	l              sync.RWMutex
 	pools          map[string]Client
 	primTopo, topo ClusterTopo
+	secondaries    map[string][]ClusterNode
 
 	closeCh   chan struct{}
 	closeWG   sync.WaitGroup
@@ -387,6 +388,7 @@ func (c *Cluster) sync(p Client) error {
 		defer c.l.Unlock()
 		c.topo = tt
 		c.primTopo = tt.Primaries()
+		c.secondaries = tt.secondariesByPrimary()
 
 		tm := tt.Map()
 		for addr, p := range c.pools {
@@ -438,6 +440,15 @@ func (c *Cluster) addrForKey(key string) string {
 	return ""
 }
 
+func (c *Cluster) secondaryAddrForKey(key string) string {
+	c.l.RLock()
+	defer c.l.RUnlock()
+	for _, node := range c.secondaries[c.addrForKey(key)] {
+		return node.Addr
+	}
+	return ""
+}
+
 type askConn struct {
 	Conn
 }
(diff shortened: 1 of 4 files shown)


nussjustin:
@mediocregopher The cluster changes are a bit more complex than the sentinel changes since I had to change the stubs a bit, but everything should work already.

nussjustin:
Changed `secondaryAddrForKey` to select a random secondary each time instead of always picking the first.

mediocregopher:
What was the status on the idea of just always setting the READONLY flag on the connection? I thought the docs you posted in #174 made it seem like there wasn't a reason to not do it, but maybe I misunderstood.

nussjustin:
The problem is that every `Client` is a `Pool`. We can't just automatically send a `READONLY` on every connection the `Cluster`. I don't see how this could be achieved, unless the user specifies a custom `ClusterPoolFunc` with then itself specifies a customer `PoolConnFunc` which then sends the `READONLY` on connection initialization.

mediocregopher:
I wonder if it'd be worth having a `DefaultClusterConnFunc`, which we could point people towards in the `ClusterPoolFunc` docs. That seems more flexible and easier to follow than telling people they have to call `READONLY` manually. What do you think?

nussjustin:
So basically tell the user "If you want to use secondary reads, pass `ClusterPoolFunc(DefaultClusterConnFunc)`" where `DefaultClusterConnFunc` would be something like

```go
var DefaultClusterConnFunc = func(network, addr string) (Client, error) {
    return NewPool(network, addr, 4, PoolConnFunc(func(network, addr string) (Conn, error) {
        c, err := DefaultConnFunc(network, addr)
        if err != nil {
            return err
        }
        if err := c.Do(Cmd(nil, "READONLY")); err != nil {
            c.Close()
            return err
        }
        return c, nil
    }))
}
```

Correct? Seems Find.

We could theoretically make `DefaultClusterConnFunc` be the default value for `ClusterPoolFunc`, but that would break for users that overwrite `DefaultClientFunc`, which I doubt any one really does.

mediocregopher:
So your `DefaultClusterConnFunc` is actually a `ClientFunc`, I was thinking of just:

```go
var DefaultClusterConnFunc = func(network, addr string) (Conn, error) {
        c, err := DefaultConnFunc(network, addr)
        if err != nil {
            return err
        } else if err := c.Do(Cmd(nil, "READONLY")); err != nil {
            c.Close()
            return err
        }
        return c, nil
}
```

That way if someone is specifying their own Client for Cluster (which is a pretty common thing to do, to mess with pool sizes and such) they don't have to worry about READONLY. We just need to make a note on `ClusterPoolFunc` about using `DefaultClusterConnFunc`.

nussjustin:
Added the `DefaultClusterConnFunc` together with an example on `ClusterPoolFunc`. Screenshot from `godoc`:

image

(btw. does the Travis integration not work? I don't see the check for Travis, but the [build still ran](https://travis-ci.org/github/mediocregopher/radix/builds/675347375))

mediocregopher:
Thanks for pushing this through and dealing with all the slow back and forth with me :)

> (btw. does the Travis integration not work? I don't see the check for Travis, but the build still ran)

I also don't see the check, that's weird. I vaguely remember github was going to deprecate hooks in favor of actions, or something like that, I wonder if they followed through with that finally.

PR #202, opened 2020-04-05
Removed excess mutex locking and provided a test for Pool.Do() (#201)

### LINKED ISSUE #201: Request timeouts are disregarded
Hi!

In my production cluster I found out that sometimes my app (HTTP API) takes much longer time to time out a request than it should. Actually with `DialTimeout` set to 225ms a single request to Redis can take up to multiple seconds:
```
2020/03/31 21:33:45.252625 error: redis: could not do action: read tcp 173.17.0.1:46346->173.17.0.4:6379: i/o timeout [4.337741181s, 1/2]
2020/03/31 21:33:45.356712 error: redis: could not do action: read tcp 173.17.0.1:46348->173.17.0.4:6379: i/o timeout [4.438620564s, 1/2]
2020/03/31 21:33:45.458909 error: redis: could not do action: read tcp 173.17.0.1:46350->173.17.0.4:6379: i/o timeout [4.537682881s, 1/2]
2020/03/31 21:33:45.561902 error: redis: could not do action: read tcp 173.17.0.1:45904->173.17.0.4:6379: i/o timeout [4.579830129s, 1/2]
```

This problem is crucial for my app since 1) goroutines in the worker pool get stuck 2) I could use that time to make a retry, but instead I do nothing and the client of my API times out.

I managed to reproduce the problem in dev cluster simply by having a large (80) number of connections per shard and filtering tcp packets with iptables to imitate timeouts. Then I narrowed down the problem to the `put` function:
```
ROUTINE ======================== github.com/mediocregopher/radix/v3.(*Pool).put in /home/kirill.alekseev/dev/becca-go/vendor/github.com/mediocregopher/radix/v3/pool.go
--
0     15.59s (flat, cum) 59.69% of Total
.          .    569:
.          .    570:   p.l.RUnlock()
.          .    571:   // the pool might close here, but that's fine, because all that's happening
.          .    572:   // at this point is that the connection is being closed
.          .    573:   ioc.Close()
.     1.46ms    574:   p.traceConnClosed(trace.PoolConnClosedReasonPoolClosed)
.          .    575:   p.l.Lock()
.          .    576:   p.totalConns--
.     15.59s    577:   p.l.Unlock()
```

Here we block on `p.l.Lock()` waiting for `p.l.RUnlock()` in `getExisting()`:
```
func (p *Pool) getExisting() (*ioErrConn, error) {
    p.l.RLock()
    defer p.l.RUnlock()
 
    ...
    select {
        case ioc := <-p.pool:
        return ioc, nil
    case <-tc:
        return nil, p.opts.errOnEmpty
    }
}
```

So if user provides `PoolOnEmptyCreateAfter` (which is my case), `p.l.` is not unlocked untill the timer fires. Meanwhile I am not sure why you need to lock the mutex for reading here since reading from a closed channel should not be a problem.

I will provide a PR soon, should not be difficult

gibsn:
Resolves #201

### DIFF
--- a/pool.go
+++ b/pool.go
@@ -493,16 +493,12 @@ func (p *Pool) doOverflowDrain() {
 }
 
 func (p *Pool) getExisting() (*ioErrConn, error) {
-	p.l.RLock()
-	defer p.l.RUnlock()
-
-	if p.closed {
-		return nil, errClientClosed
-	}
-
-	// Fast-path if the pool is not empty.
+	// Fast-path if the pool is not empty. Return error if pool has been closed.
 	select {
-	case ioc := <-p.pool:
+	case ioc, ok := <-p.pool:
+		if !ok {
+			return nil, errClientClosed
+		}
 		return ioc, nil
 	default:
 	}
@@ -523,7 +519,10 @@ func (p *Pool) getExisting() (*ioErrConn, error) {
 	}
 
 	select {
-	case ioc := <-p.pool:
+	case ioc, ok := <-p.pool:
+		if !ok {
+			return nil, errClientClosed
+		}
 		return ioc, nil
 	case <-tc:
 		return nil, p.opts.errOnEmpty
--- a/pool_test.go
+++ b/pool_test.go
@@ -236,6 +236,55 @@ func TestPoolPut(t *T) {
 	assertPoolConns(0)
 }
 
+// TestPoolDoDoesNotBlock checks that with a positive onEmptyWait Pool.Do()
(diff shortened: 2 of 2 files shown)


mediocregopher:
Hi @gibsn ! thanks for looking so deeply into this. I agree, I can't see why that RLock is needed either, probably it used to be but isn't any longer, and your fix for it looks good :) Thank you 🙏

gibsn:
@mediocregopher 🤝🤝🤝

PR #211, opened 2020-05-16
Add Empty property to MaybeNil.

### LINKED ISSUE #206: Non-working MaybeNil for hashmap keys
Currently `MaybeNil` can't be used for non-existing hashmap keys detection. Operations like `HGETALL` always returns empty list reply `$0\r\n` which treated as non-nil response.
I have two suggestions about it:
* Add such check to `MaybeNil` - can probably break compatibility
* Add `MaybeEmpty` or extend `MaybeNil` with `IsEmpty` field

xakep666:
This is useful for commands like HGETALL that always returns non-nil array responses.

Closes #206

### DIFF
--- a/action.go
+++ b/action.go
@@ -292,24 +292,31 @@ func (c *cmdAction) ClusterCanRetry() bool {
 
 // MaybeNil is a type which wraps a receiver. It will first detect if what's
 // being received is a nil RESP type (either bulk string or array), and if so
-// set Nil to true. If not the return value will be unmarshaled into Rcv
-// normally.
+// set Nil to true. If not the return value will be unmarshalled into Rcv
+// normally. If the response being received is an empty array then the EmptyArray
+// field will be set and Rcv unmarshalled into normally.
 type MaybeNil struct {
-	Nil bool
-	Rcv interface{}
+	Nil        bool
+	EmptyArray bool
+	Rcv        interface{}
 }
 
 // UnmarshalRESP implements the method for the resp.Unmarshaler interface.
 func (mn *MaybeNil) UnmarshalRESP(br *bufio.Reader) error {
 	var rm resp2.RawMessage
-	if err := rm.UnmarshalRESP(br); err != nil {
+	err := rm.UnmarshalRESP(br)
+	switch {
+	case err != nil:
 		return err
-	} else if rm.IsNil() {
+	case rm.IsNil():
 		mn.Nil = true
 		return nil
+	case rm.IsEmptyArray():
+		mn.EmptyArray = true
+		fallthrough // to not break backwards compatibility
+	default:
+		return rm.UnmarshalInto(resp2.Any{I: mn.Rcv})
 	}
-
-	return rm.UnmarshalInto(resp2.Any{I: mn.Rcv})
 }
 
 ////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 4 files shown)


action.go
```diff
@@ -295,21 +295,27 @@ func (c *cmdAction) ClusterCanRetry() bool {
 // set Nil to true. If not the return value will be unmarshaled into Rcv
 // normally.
 type MaybeNil struct {
-	Nil bool
-	Rcv interface{}
+	Nil   bool
+	Empty bool
```
mediocregopher, line 299: I'd like to name the field `EmptyArray`, to make it a bit more specific

action.go
```diff
@@ -295,21 +295,27 @@ func (c *cmdAction) ClusterCanRetry() bool {
 // set Nil to true. If not the return value will be unmarshaled into Rcv
```
mediocregopher, line 295: Can you update the comment to read as:

> MaybeNil is a type which wraps a response receiver. If the response being received is a nil RESP type (either bulk string or array), it will set Nil to true. If not the return value will be unmarshaled into Rcv normally. If the response being received is an empty array then the `EmptyArray` field will be set and Rcv unmarshaled into normally.

action_test.go
```diff
@@ -396,9 +411,12 @@ func TestMaybeNil(t *T) {
 			var rm resp2.RawMessage
 			mn := MaybeNil{Rcv: &rm}
 			require.Nil(t, mn.UnmarshalRESP(bufio.NewReader(buf)))
-			if mnt.isNil {
+			switch {
+			case mnt.isNil:
 				assert.True(t, mn.Nil)
-			} else {
+			case mnt.isEmpty:
+				assert.True(t, mn.Empty)
```
mediocregopher, line 418: The `isEmpty` case needs to also `assert.Equal(t, mnt.b, string(rm))`

resp/resp2/resp.go
```diff
@@ -1292,3 +1293,8 @@ func (rm RawMessage) UnmarshalInto(u resp.Unmarshaler) error {
 func (rm RawMessage) IsNil() bool {
 	return bytes.Equal(rm, nilBulkString) || bytes.Equal(rm, nilArray)
 }
+
+// IsEmpty returns true if the contents of RawMessage is empty array value.
+func (rm RawMessage) IsEmpty() bool {
```
mediocregopher, line 1298: Let's name this `IsEmptyArray`, to be more specific.

action.go
```diff
@@ -295,21 +295,27 @@ func (c *cmdAction) ClusterCanRetry() bool {
...
 		return err
-	} else if rm.IsNil() {
+	case rm.IsNil():
 		mn.Nil = true
 		return nil
+	case rm.IsEmpty():
+		mn.Empty = true
+		return nil
+	default:
+		return rm.UnmarshalInto(resp2.Any{I: mn.Rcv})
```
mediocregopher, line 317: `UnmarshalInto` must be performed regardless of whether or not the response is an empty array, otherwise this breaks backwards compatibility.

action.go
xakep666: I also thought about check for empty string here but in tests reading of `$0\r\n` causes `rm.UnmarshalRESP` to return EOF error.
If there is no ways to check for empty string I'll of course rename property to `EmptyArray`

mediocregopher: That's weird about the tests. In any case, I'd rather stick with just `EmptyArray` for now, as it's the minimum that we're sure we need at the moment. `EmptyBulkString`, `EmptySimpleString`, etc... can be added alongside it at a later date when it's a need that comes up.

mediocregopher:
Merged, thanks @xakep666 !

PR #218, opened 2020-07-01
fix unmarshalling StreamEntry with nil array for fields

Gaboose:
Redis allows access to pending unacknowledged messages with `XREADGROUP GROUP mygroup myconsumer STREAMS mystream 0`. If those stream entries have been trimmed or deleted, the answer will look like this:
```
1) 1) "mystream"
   2) 1) 1) "1593607546396-0"
         2) (nil)
```
`radix.StreamReader` fails to return these stream entries and returns "invalid stream entry" error instead.

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -155,16 +155,19 @@ func (s *StreamEntry) UnmarshalRESP(br *bufio.Reader) error {
 	if err := ah.UnmarshalRESP(br); err != nil {
 		return err
 	}
-	if ah.N%2 != 0 {
+
+	for k := range s.Fields {
+		delete(s.Fields, k)
+	}
+
+	if ah.N == -1 {
+		return nil
+	} else if ah.N%2 != 0 {
 		return errInvalidStreamEntry
 	}
 
 	if s.Fields == nil {
 		s.Fields = make(map[string]string, ah.N/2)
-	} else {
-		for k := range s.Fields {
-			delete(s.Fields, k)
-		}
 	}
 
 	var bs resp2.BulkString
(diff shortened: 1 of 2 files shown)


stream.go
```diff
@@ -167,6 +164,12 @@ func (s *StreamEntry) UnmarshalRESP(br *bufio.Reader) error {
 		}
 	}
 
+	if ah.N == -1 {
```
mediocregopher, line 167: I believe this if check should still come before the initialization of `Fields` above, since that's using `ah.N` to guess the map size in one of the cases. If `ah.N` is negative that would... panic? It wouldn't do anything good. Is there a reason you moved this check here?

Gaboose: Oh yeah, that'd be a problem. I moved the check down here, because I wanted to make use of this loop that clears the Fields map:
```
for k := range s.Fields {
	delete(s.Fields, k)
}
```
If `ah.N == -1` we should probably still clear the map.

Gaboose: Changed it. How's that?

Gaboose: We could also go with
```
	if s.Fields != nil {
		for k := range s.Fields {
			delete(s.Fields, k)
		}
	}

	if ah.N == -1 {
		return nil
	} else if ah.N%2 != 0 {
		return errInvalidStreamEntry
	}

	if s.Fields == nil {
		s.Fields = make(map[string]string, ah.N/2)
	}
```

mediocregopher: tbh I like this way ^ better. It's less duplicate code, and easier to read. I think the `if s.Fields != nil` isn't even necessary, ranging over a nil map is like ranging over a nil slice.

Gaboose: cool, changed

PR #220, opened 2020-07-19
Fix BenchmarkParallelGetSet results

bogdanciuca:
Fix `BenchmarkParallelGetSet/redigo` test to actually use `redigo` pipelines. See [issue](https://github.com/mediocregopher/radix/issues/221).

### DIFF
--- a/bench/bench_test.go
+++ b/bench/bench_test.go
@@ -240,11 +240,23 @@ func BenchmarkParallelGetSet(b *B) {
 
 		do(b, func() error {
 			conn := red.Get()
-			if _, err := conn.Do("SET", "foo", "bar"); err != nil {
+			if err := conn.Send("SET", "foo", "bar"); err != nil {
 				conn.Close()
 				return err
 			}
-			if out, err := redigo.String(conn.Do("GET", "foo")); err != nil {
+			if err := conn.Send("GET", "foo"); err != nil {
+				conn.Close()
+				return err
+			}
+			if err := conn.Flush(); err != nil {
+				conn.Close()
+				return err
+			}
+			if _, err := conn.Receive(); err != nil {
+				conn.Close()
+				return err
+			}
+			if out, err := redigo.String(conn.Receive()); err != nil {
 				conn.Close()
 				return err
 			} else if out != "bar" {


nussjustin:
The change is not correct, since the other 2 clients don't use manual pipelining either. The benchmark explicitly does not use pipelines for all 3 clients (redigo, redispipe, radix).

The benchmark is meant to show how implicit pipelining improves performance for concurrent goroutines. Both redispipe and radix will automatically pipeline concurrent commands under the hood as an optimization, which redigo does not do (at least it didn't the last time I checked). This allows redispipe and radix to be much faster when multiple goroutines try to run commands concurrently.

Optimizing the benchmark by using pipelines for the SET+GET may improve the performance per iteration, but doesn't make much sense considering the benchmarks goal. Also please note that (at least in my experience) people will often not use manual pipelining, even if possible and both redispipe and radix basically try to make manual pipelining for small pipelines completely unnecessary, so that people don't even have to think about it most of the time.

If we change the redigo benchmarks to use manual pipelining, then both the redispipe and radix benchmarks will have to be changed, too. Note that radix does not currently support manual pipelines with implicit pipelining (which is actually an explicit design decision to give more control over pipelining), which means that manual pipelining would probably result in a performance hit in this benchmark, but will probably in v4. redispipe also does not seem to support manual pipelining at all.

bogdanciuca:
@nussjustin I think saying that "the change is not correct" is a bit of an overstatement. I understand the codewise advantage  of radix / redispipe - which automatically pipeline commands. 

> The benchmark is meant to show how implicit pipelining improves performance for concurrent goroutines. Both redispipe and radix will automatically pipeline concurrent commands under the hood as an optimization, which redigo does not do (at least it didn't the last time I checked). This allows redispipe and radix to be much faster when multiple goroutines try to run commands concurrently.

Why is implicit pipelining the benchmarks goal? Shouldn't it be general performance and a fair comparison? The benchmarks seems to show that `redigo` doesn't support pipelining and is outperformed (which is not true, if you write a few extra lines of code). I strongly feel this is the first wrong impression most people get form these benchmarks.

nussjustin:
An extra benchmark should be ok. In fact, it'm quite interested in seeing how manual pipelining in radix compares to redigo :-)

The idea behind implicit pipelining is to reduce overhead (I/O and CPU) on both client and server without having to write manual pipelines (though even manual pipelines can benefit from implicit pipelining, just not in redispipe and radix since neither implements it). By combining commands from multiple concurrent goroutines we can batch both writes and reads, not only on the client, but also on the server, reducing the overhead of I/O (by doing less syscalls, which is by far the slowest part when using Redis) and even CPU usage.

While manual pipelining also reduces the overhead, it doesn't work as well for most cases, since both client and server still have to do the I/O and parsing per pipeline. We can easily see this when comparing the number of times the client and server have to do I/O and parsing, when comparing no pipelining with manual and implicit pipelining. For this I'm assuming 32 goroutines each doing a sequential SET + GET (similiar to the benchmark), but with only 1 iteration per goroutine:

* without pipelining (what the redigo code does) we pay 32*2 = 64 times for I/O and parsing on both client and server,
* with manual pipelining (what your modified redigo code does) we "only" pay 32 times for I/O and parsing and
* with implicit pipelining (what redispipe and radix do) in the base case we only pay 2 times for I/O and parsing, and that without any extra code!

And that is ignoring the fact, that redispipe and radix can do all that using only 1 connection, while redigo needs either multiple connections (which adds even more overhead) or needs to wait for each pipeline to finish before sending the next one.

Again we are testing what is at lest in my experience the common case of how people use Redis (many concurrent goroutines not using manual pipelining) and how redispipe and radix automatically optimize this.

The benchmark is meant to show how redispipe and radix reduce the overhead and improve the throughput and latency of your commands _for free_.

Plus since neither redispipe nor radix can use/combine manual pipelining with implicit pipelining, adding it just for redigo would make the comparison less fair (right know from a developers perspective all 3 benchmarks do the same thing, which makes them fair). Also given your stated improvement of 50%, redigo would still be much slower than redispipe and radix, thus not changing the point and only making the benchmark "unfair" in comparison.

bogdanciuca:
> An extra benchmark should be ok. In fact, it'm quite interested in seeing how manual pipelining in radix compares to redigo :-)

Thanks, I think that should work. Any recommendations where to place it?

> - without pipelining (what the redigo code does) we pay 32*2 = 64 times for I/O and parsing on both client and server,
- with manual pipelining (what your modified redigo code does) we "only" pay 32 times for I/O and parsing and
- with implicit pipelining (what redispipe and radix do) in the base case we only pay 2 times for I/O and parsing, and that without any extra code!

So I am not trying to dismiss the benefits of implicit pipelining and its simplicity of use. I also think the batching optimisations available in `radix` and `redispipe` are amazing.

> Plus since neither redispipe nor radix can use/combine manual pipelining with implicit pipelining, adding it just for redigo would make the comparison less fair (right know from a developers perspective all 3 benchmarks do the same thing, which makes them fair). Also given your stated improvement of 50%, redigo would still be much slower than redispipe and radix, thus not changing the point and only making the benchmark "unfair" in comparison.

One could argue with the fact that the benchmarks do the same thing - since the `redigo` benchmark is not using any kind pipelining. It looks comparable with `BenchmarkParallelGetSet/radix/no_pipeline-8`. 

Even though it's still slower, `redigo` rises to `~80%` of `radix/redispipe` performance.

**Current benchmarks**:

`BenchmarkParallelGetSet/radix/no_pipeline-8                        38173             28154 ns/op              67 B/op          4 allocs/op
BenchmarkParallelGetSet/radix/one_pipeline-8                      106437             11153 ns/op              69 B/op          4 allocs/op
BenchmarkParallelGetSet/radix/default-8                           104139             11192 ns/op              70 B/op          4 allocs/op
BenchmarkParallelGetSet/redigo-8                                   **43832**             26623 ns/op             118 B/op          6 allocs/op
BenchmarkParallelGetSet/redispipe-8                               116005             10135 ns/op             168 B/op          8 allocs/op`

**PR benchmarks**:

`BenchmarkParallelGetSet/radix/no_pipeline-8                        41848             25495 ns/op              67 B/op          4 allocs/op
BenchmarkParallelGetSet/radix/one_pipeline-8                      104917             11082 ns/op              70 B/op          4 allocs/op
BenchmarkParallelGetSet/radix/default-8                           102685             11121 ns/op              70 B/op          4 allocs/op
BenchmarkParallelGetSet/redigo-8                                   **85759**             13636 ns/op             118 B/op          6 allocs/op
BenchmarkParallelGetSet/redispipe-8                               113139             10025 ns/op             168 B/op          8 allocs/op`

nussjustin:
> One could argue with the fact that the benchmarks do the same thing - since the redigo benchmark is not using any kind of pipelining. It rather looks comparable with BenchmarkParallelGetSet/radix/no_pipeline-8.

The point of the benchmark is to compare the performance of doing the same thing (from the benchmarks perspective) with multiple clients. Whatever the clients do internally (e.g. implicit pipelining) doesn't and shouldn't matter to the benchmark itself and only affect the results.

As a user I don't care what optimizations my (redis) client makes. I just want it to be efficient. If I have to change my code to be more efficient just because the client doesn't do an obvious optimization, it's the clients fault.

Also while the benchmark does a simple SET+GET for which manual pipelines work quite well, it could easily be changed to something more complex with e.g. conditional commands or multiple different functions doing commands. This would still show the performance improvements from implicit pipelining (since implicit pipelining works as long as you have concurrent commands) while effectively making manual pipelining either more cumbersome (conditional commands) or even completely useless (different concurrent functions).

bogdanciuca:
> If I have to change my code to be more efficient just because the client doesn't do an obvious optimization, it's the clients fault.

Okay, I see you have very strong feelings around this part and I am not going to argue with that.

Just to try and draw to a conclusion here, could you please advice on how to place this test? Earlier in this thread you mentioned:
> An extra benchmark should be ok. In fact, it'm quite interested in seeing how manual pipelining in radix compares to redigo :-)

Could you advise on the naming and placement of the test, so it will look obvious of the **extra configuration** required to enable pipelining in `Redigo`, which would still be slower?

Thanks!

nussjustin:
> Just to try and draw to a conclusion here, could you please advice on how to place this test? Earlier in this thread you mentioned:

A simple benchmark doing the SET+GET (and maybe even more commands or even multiple sub-benchmarks with different pipeline sizes) with each client. Basically what we have now, just with manual pipelining.

> Could you advise on the naming and placement of the test, so it will look obvious of the extra configuration required to enable pipelining in Redigo (and still slower)?

I'm not sure how we could to that. In the end redigo just doesn't support implicit pipelining. While manual pipelining seems quite similiar, and both make use of Redis support for pipelining, they are just too different in my opinion. Saying that some configuration will lower the gap just doesn't seem helpful to me since a) there will still probably be quite a gap and b) depending on your application the manual pipelines may not even make a difference at all.

@mediocregopher opinions?

bogdanciuca:
> A simple benchmark doing the SET+GET (and maybe even more commands or even multiple sub-benchmarks with different pipeline sizes) with each client. Basically what we have now, just with manual pipelining.

Great, will do. I will also include `manual pipelining` in the test name.

mediocregopher:
I rewrote the bench code in the `v4` branch to test all combinations of the following:

* parallel vs serial
* large vs small key/values
* manual pipelining vs set+get
* all the different drivers

I also changed some other details, for instance the way pool size was being determined previously was resulting in unrealistically large pool sizes, so I fixed that.

I then reran all the tests in a similar environment as @nussjustin's from when he ran them (though not with as many cores). You can view the full results [here](https://github.com/mediocregopher/radix/blob/v4/bench/bench_results.txt), and I updated the README in `master` to include some highlights.

Hopefully this clears some things up, feel free to re-open this issue @bogdanciuca if you feel that there's still more to be done. Thanks!

PR #229, opened 2020-08-12
Correctly handle wrapped errors for clusters.

kixelated:
If a custom `radix.Conn` that returns wrapped errors is used in conjunction with the cluster client, then it will fail to follow MOVED/ASK redirects. This is due to `errors.As` being used incorrectly.


kixelated:
Note that I haven't tested the code. I worked around the issue by no longer wrapping errors.

mediocregopher:
Good catch, thanks @kixelated !

PR #230, opened 2020-08-19
Fixed deadlock in PersistentPubSub.cmd call after close was called

chapsuk:
Hi @mediocregopher 

It would be great if any persistentPubSub.cmd (Subscribe/Unsubscribe/etc.) calls will returns error
after pubsub was closed instead of forever lock

Thanks!

### DIFF
--- a/pubsub_persistent.go
+++ b/pubsub_persistent.go
@@ -6,6 +6,9 @@ import (
 	"time"
 )
 
+// ErrPersistentPubSubClosed returned for any cmd after PersistentPubSub was closed
+var ErrPersistentPubSubClosed = fmt.Errorf("PersistentPubSub was closed")
+
 type persistentPubSubOpts struct {
 	connFn     ConnFunc
 	abortAfter int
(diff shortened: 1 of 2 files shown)


mediocregopher:
Hello @chapsuk thanks for submitting an MR! Unfortunately I can't accept it as it currently is. Having a go-routine exist permanently for the duration of the process, even after `Close` has been called, would be considered a go-routine leak. When `Close` is called all resources held by the object, both io and go-routines, must be cleaned up.

Another solution would be to include a `closedCh` field on `persistentPubSub` which is closed by the `Close` method, and have the `cmd` method also block on that in a select, returning an error if `closedCh` has been closed.

I'm also not sure about the `ErrPersistentPubSubClosed` variable, it's very specific. I think it would be better to not expose it publicly.

Ultimately if someone is calling a method on PersistentPubSub after it's been closed that's an error on the caller's part, it would be just as acceptable for the method to panic. I agree that it just blocking forever is not ideal.

chapsuk:
Thanks for the comments! PTAL

chapsuk:
@mediocregopher let me know if I need to do something else to apply this patch)

mediocregopher:
Sorry for the delay, thanks for the fix!

Issue #228, opened 2020-08-11
Pipelining in cluster mode


mdubbyap:
I think i'm misunderstanding or just not finding the correct documentation, but trying to work with a redis cluster and do either an mget or pipelined gets. 

I see ClusterSlot, but i would have to maintain and update a slot list myself and don't really want to do this as it feels like the client should do it (and already is).

So, i figured pipelined gets should work, but they too fail with as pipelines keys are checked for the same thing as an mget is. I'm a little stumped as to how this is supposed to work. Any help would be appreciated. Thanks.

mdubbyap:
Moreover, i think the logic inside of radix.assertKeysSlot() is just wrong. It's asserting that they must be exactly the same key slot, not the same keyspace the node is handling. 

It's a shame, i really prefer the interface on this client, but it doesn't appear that cluster mode is working for anything but simple or non pipelined commands.

nussjustin:
You are right that there is no support for splitting up pipelines for clusters. But that could be implemented pretty easily by splitting up the pipelines.

One problem is that this could quickly lead to bad performance (or even worse, unpredictable performance). Another one is that you can't guarantee the order of commands which could be important.

Support this for normal non-pipelined commands would mean hardcoding special logic for some commands which can quickly get more confusing then helpful (why does it support command X but not Y? why not all commands?).

There is also an official cluster proxy that supports what you want by completely abstracting away a cluster: https://github.com/RedisLabs/redis-cluster-proxy

> Moreover, i think the logic inside of radix.assertKeysSlot() is just wrong. It's asserting that they must be exactly the same key slot, not the same keyspace the node is handling.

I think multiple keys mapping to different slots are not allowed in a single command on a Cluster, but I could be wrong about this. At least there is no way a moving slot could be handled correctly with multiple different slots in a command.

mediocregopher:
> I think multiple keys mapping to different slots are not allowed in a single command on a Cluster

This is true for transactions, a single pipeline of independent commands would work, but yeah it's not something that's reliable due to migrations like you said. @mdubbyap if you feel like you're ok with that risk then it's easy enough to wrap Pipeline's Keys method to only return a single one of the keys involved.

As for MGET, this couldn't be implemented in the normal `Cmd` because that would be too much magic. It _could_ be implemented as a new and separate action, but I never felt like implementing or supporting it. If you'd like to see it as part of this project I'd happily review an MR :)

[...]

PR #241, opened 2020-11-12
Support pipelining for cluster

larryzhu2018:
There are two changes needed to support cluster mode:
1. instead of asserting on the key slot, we can just check if the server is the same
2. we need to expose the server so that the client can group all the requests for the same server and do pipelines. If there is any key move error, the client should just try again by getting the new key server and and grouping the message accordingly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 test-tmp
+.idea
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,29 @@
 Changelog from v3.0.1 and up. Prior changes don't have a changelog.
 
+# v3.6.0
+
+**New**
+
+* Add `Typle` type, which makes unmarshaling `EXEC` and `EVAL` results easier.
+
+* Add `PersistentPubSubErrCh`, so that asynchronous errors within
+  `PersistentPubSub` can be exposed to the user.
+
+* Add `FlatCmd` method to `EvalScript`.
+
+* Add `StreamEntries` unmarshaler to make unmarshaling `XREAD` and `XREADGROUP`
+  results easier.
+
+**Fixes and Improvements**
+
+* Fix wrapped errors not being handled correctly by `Cluster`. (PR #229)
+
+* Fix `PersistentPubSub` deadlocking when a method was called after `Close`.
+  (PR #230)
+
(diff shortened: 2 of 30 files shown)


mediocregopher:
Hi @larryzhu2018, this use-case has been brought up before (issue #228 ) and I believe my response remains the same. It's a tricky use-case to fully support while handling all edge-cases (migrations, down nodes, etc...) in a clean way, so it's better left up to the user to determine what they wish to do. You can see my answer in that issue for my specific suggestion.

I'm afraid I can't accept this PR, as the functionality you need is already available and I don't wish to modify or expand radix's API for it. I hope this doesn't stop you from contributing in the future though, just please make an issue first next time so we can discuss the changes before you spend time on it :) 

Thanks!
Brian

mediocregopher:
To address your `AddrForKey` change specifically, that's probably not a good idea to use for what you're trying to do, because it's going to grab an RLock for every key. Presumably you have a lot of keys (otherwise it wouldn't be worth the effort to batch them like this), so you're better off calling Topo and grouping the keys that way.

larryzhu2018:
> Hi @larryzhu2018, this use-case has been brought up before (issue #228 ) and I believe my response remains the same. It's a tricky use-case to fully support while handling all edge-cases (migrations, down nodes, etc...) in a clean way, so it's better left up to the user to determine what they wish to do. You can see my answer in that issue for my specific suggestion.
> 
> I'm afraid I can't accept this PR, as the functionality you need is already available and I don't wish to modify or expand radix's API for it. I hope this doesn't stop you from contributing in the future though, just please make an issue first next time so we can discuss the changes before you spend time on it :)
> 
> Thanks!
> Brian

hi Brian, I read #228 before submitting this PR. as I mentioned in the description, the client should just try again by getting the new key server and and grouping the message accordingly. Does that work?

larryzhu2018:
> To address your `AddrForKey` change specifically, that's probably not a good idea to use for what you're trying to do, because it's going to grab an RLock for every key. Presumably you have a lot of keys (otherwise it wouldn't be worth the effort to batch them like this), so you're better off calling Topo and grouping the keys that way.

these are local operations, right? can you tell me how do you do that? basically I would like to group the keys for the same server. you cannot use key slots to group as that would be too many groups and not enough message to make the pipeline meaningful

PR #244, opened 2020-11-29
add new flag to StreamReader - SwitchToNewMessages

### LINKED ISSUE #243: Reading unacknowledged entries from a stream and switching to new ones
Hello!

Thanks for the `StreamReader`. It helps a lot!

I'm building a service that reads entry (1), runs some processing logic (2), acknowledges read (3), and then repeats from the first step. I want it to read history (unacknowledged entries) on startup and then switch to reading new entries. 

[Streams Intro from Redis](https://redis.io/topics/streams-intro) provides an example in Ruby for that use-case. Is that possible with the current implementation of `StreamReader`?

It would be helpful to set some `StreamReader's` flag for switching to `>` id in case of no entries available by an existing id.

Here is the example in Ruby:

```
require 'redis'

if ARGV.length == 0
    puts "Please specify a consumer name"
    exit 1
end

ConsumerName = ARGV[0]
GroupName = "mygroup"
r = Redis.new

def process_message(id,msg)
    puts "[#{ConsumerName}] #{id} = #{msg.inspect}"
end

$lastid = '0-0'

puts "Consumer #{ConsumerName} starting..."
check_backlog = true
while true
    # Pick the ID based on the iteration: the first time we want to
    # read our pending messages, in case we crashed and are recovering.
    # Once we consumed our history, we can start getting new messages.
    if check_backlog
        myid = $lastid
    else
        myid = '>'
    end

    items = r.xreadgroup('GROUP',GroupName,ConsumerName,'BLOCK','2000','COUNT','10','STREAMS',:my_stream_key,myid)

    if items == nil
        puts "Timeout!"
        next
    end

    # If we receive an empty reply, it means we were consuming our history
    # and that the history is now empty. Let's start to consume new messages.
    check_backlog = false if items[0][1].length == 0

    items[0][1].each{|i|
        id,fields = i

        # Process the message
        process_message(id,fields)

        # Acknowledge the message as processed
        r.xack(:my_stream_key,GroupName,id)

        $lastid = id
    }
end
```

ollevche:
closes #243

Hope the naming is OK :)

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -226,6 +226,12 @@ type StreamReaderOpts struct {
 	// If a value is non-nil, only newer stream entries will be returned.
 	Streams map[string]*StreamEntryID
 
+	// SwitchToNewMessages allows to read unacknowledged messages first and then switch to new ones.
+	//
+	// If SwitchToNewMessages is true and reads uses XREADGROUP, reads will switch to the special > id
+	// in case of no messages available by an exact one (it should be 0-0 initially).
+	SwitchToNewMessages bool
+
 	// Group is an optional consumer group name.
 	//
 	// If Group is not empty reads will use XREADGROUP with the Group as consumer group instead of XREAD.
@@ -285,6 +291,7 @@ func NewStreamReader(c Client, opts StreamReaderOpts) StreamReader {
 	if sr.opts.Group != "" {
 		sr.cmd = "XREADGROUP"
 		sr.fixedArgs = []string{"GROUP", sr.opts.Group, sr.opts.Consumer}
+		sr.switchToNewMessages = opts.SwitchToNewMessages
 	} else {
 		sr.cmd = "XREAD"
 		sr.fixedArgs = nil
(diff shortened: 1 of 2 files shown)


stream.go
```diff
@@ -226,6 +226,12 @@ type StreamReaderOpts struct {
 	// If a value is non-nil, only newer stream entries will be returned.
 	Streams map[string]*StreamEntryID
 
+	// SwitchToNewMessages allows to read unacknowledged messages first and then switch to new ones.
```
mediocregopher, line 229: I'd like to change the naming to `FallbackToUndelivered`, and have the comment be:

```
FallbackToUndelivered will cause any streams in with a non-nil value in Streams to fallback to delivering messages not-yet-delivered to other consumers (as if the value in the Streams map was nil), once the reader has read all its pending messages in the stream. This must be used in conjunction with Group.
```

mediocregopher:
One small comment about the naming and comment, but overall looks good, thanks for submitting :)

ollevche:
Hey @mediocregopher, I have changed the naming! Could you please accept PR?

mediocregopher:
Done! thanks for the contribution :)

Issue #245, opened 2020-12-25
Add a new cluster option


itiscold:
Hi! 
We have used radix recently and it really helps a lot. Thx for the excellent work.

In our situation, we have dozens of addresses, and some may be unuseable at the time due to problems such as unstable network. When calls NewCluster to get a Cluster instance, it will try to create connections for every address provided and return an error immediately if any one of the addresses is found to be unconnected. As a result, the NewCluter will exit with an error and our program can not start successfully. In fact, such situation is really common and the address is just temporarily unavailable. We noticed that if the pool for address is empty, radix client will try to create a new one when synchronizing or executing cmd. Besides, if some nodes really fail, redis server will take measures to ensure the availability of the cluster. So we think it is acceptable that errors of creating pool during the initialization do not cause the cluster instance fail to be created.

**What we change:**
Here we add an option called initSyncSilent to clusterOpts. If initSyncSilent is set to true, the Sync will not return an error if it fails to create pool for part of the addresses during initializing a cluster. Note that this option will only be applied to Sync when creating a new cluster instance and does not change cluster's behavior at other times.  Here is the pr: #247 

on summary: 
1. The setting we add is optional and the default is off. 
2. If initSyncSilent is set to true, it will only change the cluster's behavior during initialization. In detail, errors of creating pool for every address provided will be ignored when calls NewCluster. This can reduce the failure rate of creating a new cluster instance.
3. Even if the address is unavailable, the client will find it when calls Sync or Do, and return an error as expected. This provides a guarantee for availability.

**Test:**
In addition to the unit tests, we have conducted some experiments to verify the changes. The cluster includes 6 nodes. And connection requests to one master node from the redis client have been dropped temporarily. Then we tried to call Sync, set data to slots of the node and get data from the slots.

When the initSyncSilent was set to true, the client started successfully. But if the node was still unconnected, calls of Sync and cmds of getting and setting would get an error. On the contrary, it worked normally if the node had recovered. The results are consistent with the unit tests results.

PR #247, opened 2020-12-27
add new clusterOpt initSyncSilent

itiscold:


teou:
#245

cluster.go
```diff
@@ -331,13 +341,22 @@ func (c *Cluster) getTopo(p Client) (ClusterTopo, error) {
 // to new instances and removing ones from instances no longer in the cluster.
 // This will be called periodically automatically, but you can manually call it
 // at any time as well
-func (c *Cluster) Sync() error {
+func (c *Cluster) Sync(args ...bool) error {
```
mediocregopher, line 344: The method signature of `Sync` cannot be changed, and I think calls to `Sync` should not be subject to the silence flag at all. You will need to change the invocation of `Sync` in `NewCluster` to use `sync` instead.

cluster.go
```diff
@@ -393,7 +412,11 @@ func (c *Cluster) traceTopoChanged(prevTopo ClusterTopo, newTopo ClusterTopo) {
 
 // while this method is normally deduplicated by the Sync method's use of
 // dedupe it is perfectly thread-safe on its own and can be used whenever.
-func (c *Cluster) sync(p Client) error {
+func (c *Cluster) sync(p Client, args ...bool) error {
```
mediocregopher, line 415: `sync` should take in just one boolean to indicate if errors should be silenced, this `...bool` pattern is not idiomatic in go.

cluster_test.go
```diff
@@ -46,6 +46,98 @@ func TestClusterClient(t *T) {
...
+			defer c.l.RUnlock()
+			assert.Equal(t, c.topo, scl.topo())
+			assert.Len(t, c.pools, len(c.topo))
+			for _, node := range c.topo {
+				assert.Contains(t, c.pools, node.Addr)
+			}
+		}
+		assertClusterState()
+
+		// cluster is unstable af
```
mediocregopher, line 86: None of this has anything to do with initialization and should not be included in the test.

cluster_stub_test.go
```diff
@@ -329,6 +329,28 @@ func (scl *clusterStub) newCluster(opts ...ClusterOpt) *Cluster {
 	return c
 }
 
+func (scl *clusterStub) clientInitSyncFailedFunc(failedAddrsMap map[string]bool) ClientFunc {
+	return func(network, addr string) (Client, error) {
+		for _, s := range scl.stubs {
+			if s.addr == addr {
+				for failedAddr, flag := range failedAddrsMap {
```
mediocregopher, line 336: This would be better written as:

```go
if failedAddrsMap[s.addr] {
    failedAddrsMap[s.addr] = false
    return nil, errors.New("timeout")
}
```

cluster_test.go
```diff
@@ -46,6 +46,98 @@ func TestClusterClient(t *T) {
...
+			scl.migrateSlotRange(dstStub.addr, slotRange[0], slotRange[1])
+			assertClusterState()
+		}
+	}
+
+	//part of the addresses are unavailable during the initialization
+	//and recover after that, try Set and Get cmd to test whether it can work
+	{
+		c, err := scl.newInitSyncErrorCluster(serverAddrs,
+			makeFailedFlagMap(serverAddrs[len(serverAddrs)/2:]),
```
mediocregopher, line 113: rather than testing that half of the addresses are failing again, it'd be better to test what happens when _all_ addresses are failing (creating the cluster should fail).

cluster.go
```diff
@@ -111,6 +112,15 @@ func ClusterWithTrace(ct trace.ClusterTrace) ClusterOpt {
 	}
 }
 
+// ClusterWithInitSyncSilent tells the Cluster whether it should ignore
+// errors of creating pool for addrs during the initialization
+// if pool for all addrs failed to be created, it will still return errors
+func ClusterWithInitSyncSilent(initSyncSilent bool) ClusterOpt {
```
mediocregopher, line 118: I'd like to rename this to `ClusterOnInitAllowUnavailable`, to be more specific about what it's doing, and change the comment to:

```
ClusterOnInitAllowUnavailable tells NewCluster to succeed and not return an error as long as at least one redis instance in the cluster can be successfully connected to.
```

mediocregopher (commented):
In general I like the idea, but I think the submitted code needs quite a bit of work, and I'm not sure that it does what it's supposed to. I left a bunch of comments, let me know if there's any questions I can answer to help you along.

cluster_test.go
itiscold: Thx for the suggestion. It is essential to test what will happen if all connections fail and I will add the test. In addition, this part is to test whether the get and set commands work normally if some nodes fail during the initialization and then recover. I think this test is necessary.

cluster_test.go
itiscold: This test is to see whether the Sync can work well with the changes of sync(). Is it ok to only check whether that call of Sync() will get no error?

itiscold:
@mediocregopher  Thanks very much for your reply and review. Before further improving the submitted code, I have some questions about your comments. The questions are left in the conversation section following your comments. Looking forward to your reply and wish you a happy new year :-)

cluster_test.go
mediocregopher: Yeah, I think it's fine to only test what happens during initialization with that option set, doing this unstable test isn't worth it imo.

cluster_test.go
mediocregopher: Ok, that makes sense, thanks for explaining it.

itiscold:
@mediocregopher   I have revised the code according to your comment, please check. Thanks a lot for your careful review :-D

mediocregopher:
Looks good to me, thanks for making the modifications :)

PR #258, opened 2021-01-23
Reduce allocations for pipelines using pre-allocated slice backing memory

nussjustin:
For each `Action` added to a `Pipeline` the `Pipeline` needs to append values to at least 2 slices (`actions`, `mm` and in most cases the `Keys` slice of the `ActionProperties`) which causes a lot of allocations.

While these allocations can be avoided by reusing `Pipeline`s, newly allocated `Pipeline`s still always incur some allocations.

For a `Pipeline` with only 1 `Action` that uses 1 key a new `Pipeline` currently needs 3 extra allocations. For a `Pipeline` with 2 `Action`s each using 1 key a `new Pipeline` currently needs 6 extra allocations. Similarly a new `Pipeline` with 4 `Action`s each using 1 key currently needs 9 extra allocations.

The number of extra allocations for a `Pipeline` with N `Action`s (assuming each `Action` is for 1 key) can be calculated automatically like this: https://play.golang.org/p/fjOdoHoHpzE (Note: These numbers can change with new Go releases)

With this change new `Pipeline`s avoid some of the allocations by pre-allocating small 4-element arrays as part of the `Pipeline`s initial allocation, that are then used as initial backing memory for each of the 3 slices.

The number 4 was chosen to avoid bloating memory use too much for applications that are mainly using small `Pipeline`s (N <= 2) while still allowing up to 9 allocations to be avoided (N = 4).

```bash
$ benchstat before.txt after.txt | grep -v '/no_pipeline/'
name                                             old time/op    new time/op    delta
Drivers/serial/pipeline/small_kv/radixv4-8         28.1µs ± 3%    29.0µs ± 5%   +2.91%  (p=0.023 n=10+10)
Drivers/serial/pipeline/large_kv/radixv4-8         53.9µs ± 1%    53.7µs ± 1%     ~     (p=0.356 n=9+10)
Drivers/parallel/pipeline/small_kv/radixv4-8       4.44µs ± 1%    4.39µs ± 1%   -1.31%  (p=0.000 n=10+9)
Drivers/parallel/pipeline/large_kv/radixv4-8       16.2µs ± 0%    16.2µs ± 1%     ~     (p=0.211 n=10+9)

name                                             old alloc/op   new alloc/op   delta
Drivers/serial/pipeline/small_kv/radixv4-8          83.0B ± 0%     83.0B ± 0%     ~     (all equal)
Drivers/serial/pipeline/large_kv/radixv4-8         12.5kB ± 0%    12.5kB ± 0%     ~     (p=0.211 n=10+10)
Drivers/parallel/pipeline/small_kv/radixv4-8         422B ± 0%      502B ± 0%  +19.04%  (p=0.000 n=10+10)
Drivers/parallel/pipeline/large_kv/radixv4-8       12.9kB ± 0%    13.0kB ± 0%   +0.62%  (p=0.000 n=10+10)

name                                             old allocs/op  new allocs/op  delta
Drivers/serial/pipeline/small_kv/radixv4-8           5.00 ± 0%      5.00 ± 0%     ~     (all equal)
Drivers/serial/pipeline/large_kv/radixv4-8           5.00 ± 0%      5.00 ± 0%     ~     (all equal)
Drivers/parallel/pipeline/small_kv/radixv4-8         12.0 ± 0%       6.0 ± 0%  -50.00%  (p=0.000 n=10+10)
Drivers/parallel/pipeline/large_kv/radixv4-8         12.0 ± 0%       6.0 ± 0%  -50.00%  (p=0.000 n=10+10)
```

### DIFF
--- a/action.go
+++ b/action.go
@@ -479,9 +479,9 @@ type pipelineMarshalerUnmarshaler struct {
 // rather not expose to users.
 type pipeline struct {
 	// preallocated buffers of slices to avoid allocating for small pipelines
-	actionsBuf        [4]Action
-	mmBuf             [4]pipelineMarshalerUnmarshaler
-	propertiesKeysBuf [4]string
+	actionsBuf        [8]Action
+	mmBuf             [8]pipelineMarshalerUnmarshaler
+	propertiesKeysBuf [8]string
 
 	actions    []Action
 	mm         []pipelineMarshalerUnmarshaler
--- a/action.go
+++ b/action.go
@@ -478,6 +478,11 @@ type pipelineMarshalerUnmarshaler struct {
 // pipeline contains all the fields of Pipeline as well as some methods we'd
 // rather not expose to users.
 type pipeline struct {
+	// preallocated buffers of slices to avoid allocating for small pipelines
+	actionsBuf        [4]Action
+	mmBuf             [4]pipelineMarshalerUnmarshaler
+	propertiesKeysBuf [4]string
+
 	actions    []Action
 	mm         []pipelineMarshalerUnmarshaler
 	properties ActionProperties
@@ -489,6 +494,16 @@ type pipeline struct {
 }
 
 var _ Conn = new(pipeline)
+
+func (p *pipeline) init() {
+	p.actions = p.actionsBuf[:0]
+	p.mm = p.mmBuf[:0]
+	p.properties = ActionProperties{
+		Keys:         p.propertiesKeysBuf[:0],
+		CanPipeline:  true, // obviously
+		CanShareConn: true,
+	}
+}
 
 func (p *pipeline) reset() {
 	p.actions = p.actions[:0]
@@ -624,12 +639,9 @@ type Pipeline struct {
 
 // NewPipeline returns a Pipeline instance to which Actions can be Appended.
 func NewPipeline() *Pipeline {
-	return &Pipeline{pipeline: pipeline{
-		properties: ActionProperties{
-			CanPipeline:  true, // obviously
-			CanShareConn: true,
-		},
-	}}
+	p := &Pipeline{}
+	p.pipeline.init()
+	return p
 }
 
 // Reset discards all Actions and resets all internal state. A Pipeline with


nussjustin:
@mediocregopher WDYT?

The main reason I did this is so that it's less likely that one would need to reuse or even pool `Pipeline`s because of all the allocations. While pooling is easy, it's also easy to get wrong (e.g. introduce races) and annoying. So if we can make it so that most users don't need to think about it, that seems like a win to me.

Also as explained in the issue, the array length 4 was chosen to avoid bloating memory usage to much. I also considered a length of 8 which would avoid up to 3 more allocations. Compared to 4 this would only add 320 bytes to each `pipeline`. I hadn't measured this before so I went with 4 for the initial change, but seeing that we are talking about only 320 bytes I think using 8-element arrays would be better.

mediocregopher:
Hey @nussjustin ! You're way better at these optimizations than me so I'm glad you're back at it :)

I tried to design Pipeline so that it could be re-used, if the user cares to do that, but as long as the API doesn't have to change then I don't mind doing these optimizations also. Using arrays to pre-allocate the slices is a great trick that I really need to remember. And yeah I think 8 seems like a better number to me if it's just 320 bytes we're talking about, I know in my usages of Pipeline I go over 4 pretty often, but I'd guess I'm usually around or under 8.

WDYT of the pipeline implementation in general? It's a bit janky but it has the benefit of not relying directly on the `cmdAction` type.

nussjustin:
Increased the array size to 8. PTAL

> WDYT of the pipeline implementation in general? It's a bit janky but it has the benefit of not relying directly on the cmdAction type.

Not having (to think about) `CmdAction` just feels so good. Also as a user of radix I really like the visible changes, especially around pipelining.

While the new code (everything, not just the pipelining) feels more complex to me, that's really just because I was quite familiar with the old code.

nussjustin:
@mediocregopher ping

Seems like Travis had some problems. All tests pass on my machine.

mediocregopher:
Ah, sorry! Will merge now, thanks! I'm glad you like the new interface, it does feel a lot cleaner than before. And yeah the pipelining code in particular is pretty weird, but it works :crossed_fingers:

PR #262, opened 2021-02-25
Add a new CmdConfig for handling custom commands

### LINKED ISSUE #261: v4: Handling custom module commands
radix currently keeps 2 lists (well maps, but who cares) of commands which it consolidates to find out whether a command is 1) blocking and 2) takes any keys. By default commands (with some exceptions) are assumed to take 1 key as the first argument and to be non-blocking. 

radix uses this information to decide where to send a command in a cluster and whether a command can be safely pipelined (since pipelining requires commands not to block).

This works well with built in commands since these are all well documented and stable, but may cause problems with custom commands provided by [Redis Modules](https://redis.io/topics/modules-intro). Since radix has no special knowledge of these commands it handles them like all other, non special cased commands, assuming they 1) take 1 key and 2) can be safely pipelined.

Both assumption are not necessarily true: A custom command can interpret its arguments in any way it wants and can [block](https://redis.io/topics/modules-blocking-ops) the client if it wants to.

While probably a rare case, I'm expecting these assumptions to cause some (probably hard to debug / find) problems in the future.

The best solution would be to add support for custom commands, but I think it would also be acceptable to just document the current state for now.

Unfortunately redis itself does not expose the information needed for us to automatically handle custom commands so supporting them in radix would probably require some kind of manual registration of commands.

nussjustin:
Fixes #261

### DIFF
--- a/action.go
+++ b/action.go
@@ -233,6 +233,8 @@ func DefaultActionProperties(cmd string, args ...string) ActionProperties {
 	case noKeyCmds[cmd] || len(args) == 0:
 	case cmd == "BITOP" && len(args) > 1:
 		properties.Keys = args[1:]
+	case cmd == "MSET":
+		properties.Keys = keysFromKeyValuePairs(args)
 	case cmd == "XINFO":
 		if len(args) >= 2 {
 			properties.Keys = args[1:2]
@@ -262,6 +264,14 @@ func findStreamsKeys(args []string) []string {
 	}
 
 	return nil
+}
+
+func keysFromKeyValuePairs(pairs []string) []string {
+	keys := make([]string, len(pairs)/2)
+	for i := range keys {
+		keys[i] = pairs[i*2]
+	}
+	return keys
 }
 
 func (cfg CmdConfig) actionProperties(cmd string, args ...string) ActionProperties {
--- a/action.go
+++ b/action.go
@@ -191,9 +191,11 @@ func (c *cmdAction) String() string {
 	return "[" + strings.Join(ss, " ") + "]"
 }
 
-// CmdConfig allows creating Actions for commands using custom logic for
-// determining the ActionProperties without the need for creating a custom
-// Action implementation.
+// CmdConfig is used to create redis command Actions with particular settings. All
+// fields are optional, all methods are thread-safe.
+//
+// The global Cmd and FlatCmd functions are shortcuts for using an empty
+// CmdConfig{}.
 //
 // This can be useful for working with custom commands provided by Redis
 // modules for which the built in logic may return sub-optimal properties or in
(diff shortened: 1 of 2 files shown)


nussjustin:
@mediocregopher This is my proposed implementation for the `CmdConfig` type you described in #261.

While writing this I also realized that we don't actually handle `MSET` correctly right now, since our default logic only finds the first key and not all keys. This should probably be fixed in a separate change (new case for `defaultActionProperties` or maybe even splitting `defaultActionProperties` into multiple different functions and calling them based on the called command).

action.go
```diff
@@ -280,6 +191,173 @@ func (c *cmdAction) String() string {
 	return "[" + strings.Join(ss, " ") + "]"
 }
 
+// CmdConfig allows creating Actions for commands using custom logic for
```
mediocregopher, line 194:
```
CmdConfig is used to create redis command Actions with particular settings. All
fields are optional, all methods are thread-safe.

The global Cmd and FlatCmd functions are shortcuts for using an empty
CmdConfig{}.
```

No need to document ActionProperties on the global level, that can be done on the field.

action.go
```diff
@@ -280,6 +191,173 @@ func (c *cmdAction) String() string {
...
+	// used to set the ActionProperties for the new Action.
+	//
+	// If ActionProperties is nil, a builtin default callback will be used that
+	// should work with all standard Redis commands but may not return correct
+	// results for custom commands provided by Redis Modules or unreleased
+	// commands.
+	ActionProperties func(cmd string, args ...string) ActionProperties
+}
+
+func defaultActionProperties(cmd string, args ...string) ActionProperties {
```
mediocregopher, line 217: I think it might be useful to make this public, so someone's custom implementation can fallback to it if `cmd` isn't one of the custom commands they're expecting.

action.go
```diff
@@ -280,6 +191,173 @@ func (c *cmdAction) String() string {
...
+
+func (cfg CmdConfig) actionProperties(cmd string, args ...string) ActionProperties {
+	pf := cfg.ActionProperties
+	if pf == nil {
+		pf = defaultActionProperties
+	}
+	return pf(cmd, args...)
+}
+
+// Cmd works like the global Cmd function but using the value returned by
```
mediocregopher, line 268: I don't think it's necessary to re-document ActionProperties here, this could simple read:

```
Cmd works like the global Cmd function but can be additionally configured using
fields on CmdConfig. See the global Cmd's documentation for further details.
```

and similarly for FlatCmd

action.go
```diff
@@ -280,6 +191,173 @@ func (c *cmdAction) String() string {
...
+//
+// The ActionProperties for the Action method will be automatically determined
+// based on the command used. For non-builtin commands (e.g. provided by Redis
+// Modules) the ActionProperties may not always return correct results.
+//
+// In this case it's possible to create a new CmdConfig instance with a custom
+// ActionProperties resolver and creating the Action via the CmdConfig`s Cmd
+// method.
+//
+// See the documentation on CmdConfig for more information.
```
mediocregopher, line 329: I think the previous two paragraphs (`The ActionProperties for the Action method...` and `In this case...`) could be omitted, and this line could just say `See CmdConfig's documentation if more configurability is required, e.g. if using commands provided by Redis Modules`.

Same with on FlatCmd.

action_test.go
```diff
@@ -243,6 +243,104 @@ func ExampleFlatCmd_struct() {
...
+}
+
+// ExampleCmdConfig_Cmd shows how to use the CmdConfig type to handle an
+// imaginary CUSTOM.MZADD command which allows adding members to multiple
+// sorted sets at once.
+//
+// The assumed syntax for the command is:
+//
+// 	CUSTOM.MZADD key score member [key score member ...]
+func ExampleCmdConfig() {
```
mediocregopher, line 308: Thank you for adding this :pray:

mediocregopher (changes requested):
Thanks Justin!

Left some suggestions around the documentation. In general we want to reduce repetition in the docs, and instead have the docs refer to each other in some sensible way. The hard part is prediction which order people will general discover the docs in; in this case I think we can expect that people will usually land on the global `Cmd`/`FlatCmd` first, so those can keep the general usage documentation for commands. People with more specific needs will get directed from there to `CmdConfig`, and then `ActionProperties` if that's what they're looking for.

Thanks for pointing out the issue with MSET keys as well, if you'd like to fix that as part of this I won't mind, or I can do it after this is merged.

action.go
nussjustin: Done

action.go
nussjustin: Done

action.go
nussjustin: Done

action.go
nussjustin: Done

nussjustin:
PTAL

mediocregopher:
Merged, thanks!

PR #268, opened 2021-03-21
v4: Allow setting the protocol version for new connections

### LINKED ISSUE #266: v4: Run tests with RESP 3 enabled
Currently all tests use the default RESP protocol version RESP 2.

RESP 3 has a lot of changes which can cause problems with commands that expect to receive responses of specific types.

I patched `conn.go` to always use RESP 3 like this:

```diff
diff --git a/conn.go b/conn.go
index 97f63ac..ea94d86 100644
--- a/conn.go
+++ b/conn.go
@@ -334,6 +334,11 @@ func (d Dialer) Dial(ctx context.Context, network, addr string) (Conn, error) {
                conn.writer(ctx, d.WriteFlushInterval)
        })
 
+ if err := conn.Do(ctx, Cmd(nil, "HELLO", "3")); err != nil {
+         conn.Close()
+         return nil, err
+ }
+
        if d.AuthUser != "" {
                if err := conn.Do(ctx, Cmd(nil, "AUTH", d.AuthUser, d.AuthPass)); err != nil {
                        conn.Close()
```

This results in a lot of errors for pubsub and streams related logic which do not happen with RESP 2. These are obviously bugs in radix and should be fixed before the v4 release.

As part of fixing these bugs and preventing future regressions I propose that we

1. Add a new `Protocol` field to `Dialer` which allows to set the RESP version used for new connections.
2. Add a new flag / environment variable that enables RESP 3 for all test connections by setting the `Protocol` field for `dial`.

nussjustin:
Also fixes the unmarshaling logic for pubsub and stream responses which failed to handle RESP3 response types and adds RESP3 to the tests run by Travis.

Fixes #266

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"net"
 	"net/url"
-	"strconv"
 	"strings"
 	"time"
 
@@ -192,9 +191,9 @@ type Dialer struct {
 
 	// Protocol can be used to automatically set the RESP protocol version.
 	//
-	// If Protocol is not 0 the Dialer will send a HELLO command with the
+	// If Protocol is not empty the Dialer will send a HELLO command with the
 	// value of Protocol as version, otherwise no HELLO command will be send.
-	Protocol int
+	Protocol string
 
 	// NetDialer is used to create the underlying network connection.
 	//
@@ -341,8 +340,8 @@ func (d Dialer) Dial(ctx context.Context, network, addr string) (Conn, error) {
 		conn.writer(ctx, d.WriteFlushInterval)
 	})
 
-	if d.Protocol != 0 {
-		args := []string{strconv.Itoa(d.Protocol)}
+	if d.Protocol != "" {
+		args := []string{d.Protocol}
 		if d.AuthUser != "" {
 			args = append(args, "AUTH", d.AuthUser, d.AuthPass)
 		} else if d.AuthPass != "" {
--- a/conn.go
+++ b/conn.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net"
 	"net/url"
+	"strconv"
 	"strings"
 	"time"
 
@@ -189,6 +190,12 @@ type Dialer struct {
 	// database index set, this takes precedence.
 	SelectDB string
 
+	// Protocol can be used to automatically set the RESP protocol version.
+	//
+	// If Protocol is not 0 the Dialer will send a HELLO command with the
+	// value of Protocol as version, otherwise no HELLO command will be send.
+	Protocol int
+
 	// NetDialer is used to create the underlying network connection.
 	//
 	// Defaults to net.Dialer.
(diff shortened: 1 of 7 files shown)


nussjustin:
@mediocregopher I noticed tests are only running against a single Go version with v4. Going forward, will radix continue to only test against the latest version or will the tests go back to testing the last 2 versions? The current Go version is 1.16 so we should at least change Travis to use this.

mediocregopher:
Hey @nussjustin ! Just wanted to let you know I haven't forgotten about your PRs, they've been sitting in my inbox but I just haven't had the chance to really look at them yet. I will try to this weekend. Sorry it's taking me so long -_-

conn.go
```diff
@@ -189,6 +190,12 @@ type Dialer struct {
 	// database index set, this takes precedence.
 	SelectDB string
 
+	// Protocol can be used to automatically set the RESP protocol version.
+	//
+	// If Protocol is not 0 the Dialer will send a HELLO command with the
+	// value of Protocol as version, otherwise no HELLO command will be send.
+	Protocol int
```
mediocregopher, line 197: This may be too much flexibility but part of me feels like this would be better as a string, in case there's some shenanigans with like a `3.5` or `4-beta` protocol in the future (or something like that). Also radix might not be connecting to redis, but to some other server that looks like redis (eg keydb), and might do its own weirdness for the protocol. wdyt?

resp/resp3/resp.go
```diff
@@ -144,7 +143,8 @@ func discardMulti(br resp.BufferedReader, l int, o *resp.Opts) error {
 	return nil
 }
 
-func discardAttribute(br resp.BufferedReader, o *resp.Opts) error {
+// DiscardAttribute discards a RESP3 attribute if there is one.
```
mediocregopher, line 146: Let's make the comment:

> DiscardAttribute discards the next RESP3 message if it is an attribute message. If the next message is not an attribute message then DiscardAttribute does nothing.

resp/resp3/resp.go
```diff
@@ -158,6 +158,15 @@ func discardAttribute(br resp.BufferedReader, o *resp.Opts) error {
 	return discardMulti(br, attrHead.NumPairs*2, o)
 }
 
+// HasPrefix returns true if the next value in the given reader has the given
```
mediocregopher, line 161: I'm somewhat uneasy about this not returning the error. I guess in most cases the error will probably get returned by the next read operation which is performed, but I think there's still opportunity for weird edge cases where an error doesn't get properly propagated at the point it actually occurred. Would it be a giant pain to have this return the error if there is one?

Some other nitpicking: `HasPrefix` is a bit ambiguous, wdyt about `NextMessageIs`, or something like that?

mediocregopher (changes requested):
Overall LGTM, just some small nitpicks about naming and such :pray:

conn.go
nussjustin: Makes sense. Done.

resp/resp3/resp.go
nussjustin: Done

I also checked all current uses of `HasPrefix`/`NextMessageIs` but in all cases the error can be ignored, so I did not add any extra error handling.

nussjustin:
Sometimes I really hate Travis :-(

Maybe it would make sense to split the tests into multiple jobs? For example one for RESP 2 and one for RESP 3 plus one extra job for linting. This would reduce the number of timeouts we get. Also I just noticed we are still on Go 1.15.

mediocregopher:
Yeah let's just add the `-v` flag. We should probably switch to github actions anyway, it seems like travis has become an afterthought...

nussjustin:
Wuhu! It passed!

mediocregopher:
So it did, thanks justin!

Issue #277, opened 2021-05-26
v4: Incorrect data returned by shared connection in presence of errors


apersson:
Requests sent over a shared connection in a connection pool can return incorrect data if earlier requests experienced errors (e.g., timeouts)

If a request in `conn.reader()` returns before consuming its response from the receive buffer, then the buffer is left in an inconsistent state. While the connection is eventually closed due to the error, there can be in-flight requests that are processed before then, and the responses for those requests will be read from the now inconsistent receive buffer.

For example, if there are 3 connMarshalerUnmarshaler's (R1, R2, R3) in rCh and the context for R1 has timed out. R1 sees an error right away and the receive buffer is untouched. R2 follows and reads the buffer and consumes R1's response, and R3 would get R2's response.

Similarly, on the write-path, if a write failed, then the receive buffer can also get into an inconsistent state as it's not known how much of the buffer was successfully flushed and which responses will be received.

nussjustin:
If a write fails, all other writes have to be either canceled or executed on another connection.

If we separated marshaling from writing to the connection (e.g. by first marshaling to an in memory buffer) we could skip commands that failed to marshal and only cancel the other commands when we got an error from writing. This would be a small hit to performance if we do this on the writer goroutine because of the copying but should not be too noticeable (and theoretically we could even do this before sending the action to the write channel, thus potentially even reducing the time needed for marshaling since all commands would be marshaled concurrently, but that would be a more complicated).

The reader part is similar to #274, but unlike #274 this is also about what happens when we timeout / get an error before even reading from the connection.

For reads the only thing we can do is to read and discard the response that would have been unmarshaled after sending the previous error to the goroutine that executed the command. This could mean waiting a long time but I don't see any other way to handle this.

apersson:
Thanks for the quick response, @nussjustin. I've linked in a potential fix.

mediocregopher:
Thank you both @apersson and @nussjustin for looking into this. I think the solution to this problem will be related and dependent on the solution to #274, so I'd like to resolve that before diving into this one.

mediocregopher:
@apersson can you retest now that #274's PR is merged? I believe it should address this issue.

apersson:
@mediocregopher, it resolved it for the test application I used to reproduce the issue. We are still seeing data mismatches in the context of larger application -- but that could be on our side. I'll file a new issue if I'm able to trace something back to radix.

apersson:
And thanks for the quick fix.

[...]

PR #278, opened 2021-05-26
v4: Incorrect data returned by shared connection in presence of errors

apersson:
Addresses #277. There are several optimizations that can be done, but the goal here is to address the correctness issue.


mediocregopher:
This was fixed as part of #279.

Thanks for the PR anyway @apersson :pray:

PR #279, opened 2021-05-27
Make EncodeDecode wait for the reader when the context is canceled.

### LINKED ISSUE #274: v4: Potential for messages to be lost due to context deadline
### Background

This bug primarily affects subscription based use-cases, such as pubsub or streams.

The following sequence of events can happen:

1) [EncodeDecode](https://github.com/mediocregopher/radix/blob/bcfb640342e3115c83b77c0d7833f82b33f28210/conn.go#L117) is called on `Conn` with a context that has some kind of deadline.

2) The command information gets passed through the codepath until it reaches the `reader` within `Conn`, [here](https://github.com/mediocregopher/radix/blob/bcfb640342e3115c83b77c0d7833f82b33f28210/conn.go#L70).

3) [Unmarshal](https://github.com/mediocregopher/radix/blob/bcfb640342e3115c83b77c0d7833f82b33f28210/conn.go#L85) is called, and blocks.

4) At the same moment, the call to `Unmarshal` successfully reads a message, and the context is canceled.

5) Before the `reader` can send back a message to the response channel [here](https://github.com/mediocregopher/radix/blob/bcfb640342e3115c83b77c0d7833f82b33f28210/conn.go#L96), the original call to `EncodeDecode` unblocks on the [doneCh](https://github.com/mediocregopher/radix/blob/bcfb640342e3115c83b77c0d7833f82b33f28210/conn.go#L136), because the context was canceled.

The result is that `context.DeadlineExceeded` or `context.Canceled` is returned from `EncodeDecode`, even though a message was successfully read from the connection. That message is simply dropped.

For most use-cases this isn't a big deal, but for something like pubsub where it's expected for messages to be infrequent and context deadline exceeding to happen normally, it will almost definitely result in messages being randomly dropped.

This bug also happens to affect the `Stub`, so whatever fix is applied here must get applied there as well.

### Solution

The `Conn` needs to be reworked no matter what. It's the biggest bottleneck of radix due to all the channel read/writes. I've taken a stab at this a couple of times but it hasn't worked out yet.

It'd be nice if golang provided a way to cancel a read off a connection with a context, because the lack of being able to do that is really what's hurting us here. The alternative seems to be something along the lines of storing the dropped message for the next read, and somehow signaling back to the reader that the message was indeed dropped.

mediocregopher:
If EncodeDecode doesn't do so then it can't be known that the response
was actually successfully read just before the context was canceled, and
so the response would just be dropped. This is especially probelematic
in the case of a pubsub reader.

Fixes #274

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -38,6 +38,41 @@ type Conn interface {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+type wrappedNetConn struct {
+	net.Conn
+	bytesRead int
+	addr      net.Addr
+}
+
+var _ net.Conn = new(wrappedNetConn)
+
+func (c *wrappedNetConn) Read(b []byte) (int, error) {
+	n, err := c.Conn.Read(b)
+	c.bytesRead += n
+
+	if err == nil || !errors.Is(err, os.ErrDeadlineExceeded) || n == 0 {
+		return n, err
+	}
+
+	// a timeout was reached, but there were bytes read before it was reached.
+	// In that case we pretend no timeout was reached. If there's more data to
+	// be pulled then we can do so, but if there's not more data to be pulled
+	// then a timeout will be returned the subsequent call to Read.
+	err = c.Conn.SetReadDeadline(time.Now().Add(500 * time.Millisecond))
+
+	return n, err
+}
+
+func (c *wrappedNetConn) resetBytesRead() {
+	c.bytesRead = 0
+}
+
+func (c *wrappedNetConn) RemoteAddr() net.Addr {
+	return c.addr
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
 type connMarshalerUnmarshaler struct {
 	ctx                    context.Context
 	marshal, unmarshalInto interface{}
@@ -48,7 +83,7 @@ type connMarshalerUnmarshaler struct {
 type conn struct {
 	proc *proc.Proc
 
-	conn         net.Conn
+	conn         *wrappedNetConn
 	rOpts, wOpts *resp.Opts
 	br           resp.BufferedReader
 	bw           resp.BufferedWriter
@@ -77,6 +112,7 @@ func (c *conn) Close() error {
 }
 
 func (c *conn) reader(ctx context.Context) {
+	var discard uint
 	doneCh := ctx.Done()
 	for {
 		select {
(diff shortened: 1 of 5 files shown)


conn.go
```diff
@@ -35,21 +38,70 @@ type Conn interface {
...
 
+type wrappedNetConn struct {
+	net.Conn
+	bytesRead int
+	addr      net.Addr
+}
+
+var _ net.Conn = new(wrappedNetConn)
+
+func (c *wrappedNetConn) Read(b []byte) (int, error) {
```
mediocregopher, line 49: The idea behind this wrapper is that the bufio.Reader will make subsequent calls to Read (within `resp3.Unmarshal`) until it has all the data it needs for a particular message to be read. Since this wrapped Read can extend its read deadline automatically it greatly reduces the chances of getting partial reads, which we can't recover from.

conn.go
```diff
@@ -62,36 +112,90 @@ func (c *conn) Close() error {
...
+					if mu.marshal != nil {
+						discard++
+					}
+
+				} else {
+					// if Unmarshal returned a context error but data was also
+					// read it means that a message was only partially read off
+					// the wire. The Conn is unusable at this point, close it
+					// and bail.
+					go c.Close()
```
nussjustin, line 193: I can't help but think that it would be nicer if a goroutine started by a `*Proc` could just return an error and have the `*Proc` automatically clean up everything (either using a per-proc cleanup method or per-goroutine like the [github.com/oklog/run package](https://github.com/oklog/run#examples))

conn.go
```diff
@@ -62,36 +112,90 @@ func (c *conn) Close() error {
...
+
+			// if discarding didn't fail then read the actual desired response.
 			if err != nil {
-				// simplify things for the caller by translating network
-				// timeouts into DeadlineExceeded, since that's actually what
-				// happened.
-				var netErr net.Error
-				if errors.As(err, &netErr) && netErr.Timeout() {
-					err = context.DeadlineExceeded
+				err = resp3.Unmarshal(c.br, mu.unmarshalInto, c.rOpts)
```
nussjustin, line 167: I can't comment the line above since it was not changed (thanks Github...) but the check should be `if err == nil` and not `if err != nil`

nussjustin (changes requested):

conn.go
mediocregopher: hmmm yeah I don't disagree, I think proc could be made to work that way fairly easily

conn.go
mediocregopher: ah yeah, good catch :bow:

mediocregopher:
Alright, I wrote up some better tests around the non-pubsub case and made everything work there. I took out the write deadline logic because a) it's not really needed, writing to redis should never really block, so a sane timeout should be fine, and b) it wasn't working correctly (writes were just immediately returning with a timeout error, even though the timeout hadn't passed yet??), and getting it to work alongside the write+flush mechanics was being weird.

nussjustin:
SGTM, now you only need to make Travis happy

conn_writer.go
```diff
@@ -114,13 +105,27 @@ func (cw *connWriter) flush() bool {
...
+	// going to check on it anyway. The write method does a basic check at the
+	// beginning to make sure that the context isn't already canceled, but
+	// that's it.
+	if err := cw.conn.SetWriteDeadline(time.Now().Add(2 * time.Second)); err != nil {
+		err = fmt.Errorf("setting write deadline prior to flush: %w", err)
 		for _, mu := range flushBuf {
 			mu.errCh <- err
 		}
+
+	} else if err := cw.bw.Flush(); err != nil {
```
apersson, line 123: If the Flush fails, should we not close the connection here to ensure that no subsequent request is written and forwarded to the reader?

mediocregopher: I'm open to that, but it would need to do so if the write fails as well. I think we should leave that out of this PR for now, and implement it alongside justin's suggestion of using something like github.com/oklog/run for go-routine management. The way go-routine management is done now would make doing this tricky, but with that package's pattern it would be a lot easier.

mediocregopher:
> SGTM, now you only need to make Travis happy

Story of my life

mediocregopher:
I can't make travis happy, but tests pass locally so I'm happy with it. I'm reading up on github actions so we can switch to that, travis seems to be a dying project.

PR #285, opened 2021-06-21
v4: reader can leave read buffer in an inconsistent state

apersson:
Addresses inconsistent issue in #277.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -206,7 +206,7 @@ func (c *conn) reader(ctx context.Context) {
 					mu.errCh <- fmt.Errorf("after partial read off Conn: %w", err)
 					return
 				}
-			} else if err != nil {
+			} else if err != nil && !errors.As(err, new(resp.ErrConnUsable)) {
 				go c.Close()
 				mu.errCh <- fmt.Errorf("unexpected error on Conn: %w", err)
 				return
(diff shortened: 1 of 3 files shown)


mediocregopher:
Thanks @apersson ! Great job catching this bug, and thank you for submitting the fix :) One thing before merging: it looks like the ` TestConnDeadlineExceeded` test is no longer passing on your branch (at least, on my local machine). I'm not entirely sure why, I might be able to look into it more soon if you're not able to, but just wanted to point that out as being a blocker before merging.

(Also I'm not sure why the Github Action to do the test didn't run here.... it definitely should have)

apersson:
It's because the test is now hitting the 'partial read' error and the connection is closed. I've updated the test so that it starts with a high timeout (to ensure success) and then reduce it until it hits a timeout.

mediocregopher:
Nice, merged! Hopefully that fixes the last of those inconsistencies you were seeing :pray:

apersson:
Thanks. The number of inconsistencies have dropped further, but are unfortunately still present. I'm going to do a test where the write path is closed right away under all error conditions to see if it resolves the remaining inconsistencies.

However, while testing these changes I encountered another issue: the sequence number logic on the read path does not work with `radix.Pipeline`.

Issue #290, opened 2021-07-08
MaybeNil has incorrect Nil value


daniel-santos:
When using the CLI, If I `SET` a key to an empty string then run a `GET` on it, the result is an empty string.  If I run `GET` on a non-existent key, then the result is nil.

```
192.168.0.8:22538> flushall
OK
192.168.0.8:22538> set jack ""
OK
192.168.0.8:22538> get jack
""
192.168.0.8:22538> get jill
(nil)
```

However when I call "get jack" from radix using a `MaybeNil`, the Nil field is incorrectly set to true.

```
var repStr string
rep := radix.MaybeNil {Rcv: &repStr}
radix.FlatCmd(&rep, "get", "jack")
if rep.Nil {
    panic("uh oh")
}
```

For this case, at least, it seems to break the entire purpose of `MaybeNil`.  I'm using Redis 6.2.3 on Gentoo and the `master` branch of radix.

```
$ redis-server --version
Redis server v=6.2.3 sha=00000000:0 malloc=jemalloc-5.2.1 bits=64 build=cd7f3369da8b56f0
```

mediocregopher:
@daniel-santos the tests for MaybeNil cover this pretty well,, and if I try to repro using your code I also don't see the problem:

```go
var repStr string
rep := radix.MaybeNil {Rcv: &repStr}
radix.FlatCmd(&rep, "get", "jack")
if rep.Nil {
    panic("uh oh")
}
```

I see in your fix on your fork that you're resetting the values of `Nil` and `EmptyArray`, which tells me maybe you're re-using instances of `MaybeNil`? That's not the expected way to use the type, so it would make sense it doesn't work right.

daniel-santos:
I'm so sorry for responding late.

>  see in your fix on your fork that you're resetting the values of `Nil` and `EmptyArray`, which tells me maybe you're re-using instances of `MaybeNil`? That's not the expected way to use the type, so it would make sense it doesn't work right.

Yes, indeed that was the problem.  I suppose I can see some other challenges with doing it that way, because you may not change the value pointed to by Rcv when it is Nil == true, but that should just be a matter of examining the value of Nil and/or EmptyArray before relying upon the value of Rcv (which could have been inited on a previous call.

Since Nil and EmptyArray should always be set, I think that my proposed patch should be safe and correct in all cases.  Thank you!

PR #291, opened 2021-07-08
Fix for #290: MaybeNil has incorrect Nil value

daniel-santos:
In my case, I'm re-using a `MaybeNil` object that had previously had it's `Nil` member set to true.  The solution is to always set `Nil` and `EmptyArray` when used as a receiver so that an incorrect result is never returned.


daniel-santos:
oops, typoed the Issue number

mediocregopher:
Thanks @daniel-santos , I didn't see this when I responded to the Issue. Re-using MaybeNil's isn't really intended to be supported, but I'll go ahead and merge this since it doesn't really hurt anything.

daniel-santos:
@mediocregopher Thank you.

PR #292, opened 2021-07-17
Properly return errors from Pipeline, deal with discard in Conn better

### LINKED ISSUE #287: v4: Pipeline no longer work with Pool
Responses are easily reordered when using `Pipeline` on a `Pool` in the presence timeouts. I believe this happens because the `discard` counter in `reader()` doesn't take into account that `Pipeline` would need to discard multiple responses.

mediocregopher:
Pipeline was not properly collecting and returning the errors it
encountered, which prevented Conn from properly discarding returned
messaged after timeout errors.

Additionally, Conn wasn't properly discarding messages. It now uses the
unmarshaler given to it to do the discard, rather than discarding a
single message. This allows for cases like pipeline where we actually
need to discard multiple messages.

Fixes #287

### DIFF
--- a/action.go
+++ b/action.go
@@ -594,12 +594,6 @@ func (p *pipeline) EncodeDecode(_ context.Context, m, u interface{}) error {
 	return nil
 }
 
-func (p *pipeline) setErr(startingAt int, err error) {
-	for i := startingAt; i < len(p.mm); i++ {
-		p.mm[i].err = err
-	}
-}
-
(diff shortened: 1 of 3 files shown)


mediocregopher:
@apersson can you check this out and let me know if it fixes the pipeline timeout issue for you? I updated the test and was able to repro out-of-order reads, but I'd like to be sure there's not some further bug that's only triggered in your setup.

mediocregopher:
@apersson I've gone ahead and merged this, and will close the original issue. Feel free to re-open the original issue if this fix isn't complete.

Issue #289, opened 2021-07-06
may add maxlifetime in the connection pool？


fare83:
in my project, redis addr use domain name, and domain name lookup ip  will be change  in some times。may add maxlifetime in the connection pool like database/sql ?  it can help for auto reconnect in some scenarios like nslookup result change。thank you。

mediocregopher:
Hi @fare83 , I would be interested in adding this feature, would you be able to implement a PR for it?

fare83:
Hi @mediocregopher, I pull a PR for it. Please code review for it, thank you.

PR #294, opened 2021-07-19
add max life time in pool

fare83:
#289

### DIFF
--- a/pool.go
+++ b/pool.go
@@ -526,7 +526,7 @@ func (p *Pool) getExisting() (*ioErrConn, error) {
 			}
 			if ioc.expired(p.opts.maxLifetime) {
 				ioc.Close()
-				p.traceConnClosed(trace.PoolConnClosedReasonPoolFull)
+				p.traceConnClosed(trace.PoolConnClosedReasonConnExpired)
 				atomic.AddInt64(&p.totalConns, -1)
 				continue
 			}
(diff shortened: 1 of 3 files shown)


pool.go
```diff
@@ -497,21 +517,6 @@ func (p *Pool) doOverflowDrain() {
 }
 
 func (p *Pool) getExisting() (*ioErrConn, error) {
-	// Fast-path if the pool is not empty. Return error if pool has been closed.
```
mediocregopher, line 500: We need to keep this fast path here, we don't want to have to create/fetch a new timer on every command as that adds extra overhead.

You should be able to abstract that for loop into a new method though, which accepts a timer channel which can be nil, and then use that method in both places we're currently selecting on `p.pool` here.

PR looks great otherwise though, thanks for doing this!

pool.go
```diff
@@ -67,6 +70,13 @@ func (ioc *ioErrConn) Close() error {
 	return ioc.Conn.Close()
 }
 
+func (ioc *ioErrConn) expired(timeout time.Duration) bool {
+	if timeout <= 0 {
+		return false
+	}
+	return ioc.createdAt.Add(timeout).Before(time.Now())
```
nussjustin, line 77:
```suggestion
	return time.Since(ioc.createdAt) >= timeout
```

pool.go
```diff
@@ -498,13 +518,21 @@ func (p *Pool) doOverflowDrain() {
...
-			return nil, errClientClosed
+	for {
+		select {
+		case ioc, ok := <-p.pool:
+			if !ok {
+				return nil, errClientClosed
+			}
+			if ioc.expired(p.opts.maxLifetime) {
+				ioc.Close()
+				p.traceConnClosed(trace.PoolConnClosedReasonPoolFull)
```
mediocregopher, line 529: We should add a new trace reason, `trace.PoolConnClosedReasonMaxLifetime`, this current trace isn't accurate (and the one in the below for loop as well).

fare83:
hi, @mediocregopher  is there other suggest for this PR ？

mediocregopher:
Sorry about the @fare83, it fell out of my inbox. I went ahead and merged it, thanks for the contribution!

PR #297, opened 2021-09-30
fix an endless loop when pool get in fastpath

warriorguo:
There is no way out when trying to get an existing connection from pool in the fast path.
This commit adds a line to break the loop when failed to obtain on the try.


mediocregopher:
Ah good catch, thank you!

Issue #296, opened 2021-08-05
Why doesn't 'saneDefault' handle error response? 


shrimptttttttttt:
In radix do command, I pass a *interface{} to the Cmd's rcv
```
func Cmd(rcv interface{}, cmd string, args ...string) CmdAction {
	c := getCmdAction()
	*c = cmdAction{
		rcv:  rcv,
		cmd:  cmd,
		args: args,
	}
	return c
}
```

Server returns an error response starts with an error prefix, namely '-', then we receive a panic as follows
```
func saneDefault(prefix byte) interface{} {
	// we don't handle errPrefix because that always returns an error and
	// doesn't touch I
	switch prefix {
	case arrayPrefix[0]:
		ii := make([]interface{}, 8)
		return &ii
	case bulkStrPrefix[0]:
		bb := make([]byte, 16)
		return &bb
	case simpleStrPrefix[0]:
		return new(string)
	case intPrefix[0]:
		return new(int64)
	}
	panic("should never get here")
}
```

My question is what's the reason of this design?
I expect to receive an error response, rather than process panics.

mediocregopher:
Hi @shrimptttttttttt this is not an intentional design decision, simply a hole in the logic which needs to be fixed, thanks for bringing it to my attention :) I will come up with a fix in the next week or so (unless someone beats me to it 😉 )

mediocregopher:
Well weeks became months, but in any case I was able to come back to this to investigate. @shrimptttttttttt I'm not able to reproduce any panic by passing an `*interface{}` into an erroring `Cmd` on either `v3` or `v4`. Here's the test cases I tried:

**v3**:

```go
func TestBlah(t *testing.T) {
    c := dial()

    var into interface{}
    err := c.Do(Cmd(&into, "UNKNOWN", "COMMAND"))

    var respErr resp2.Error
    ok := errors.As(err, &respErr)

    assert.True(t, ok)
    assert.True(t, strings.HasPrefix(respErr.E.Error(), "ERR unknown command `UNKNOWN`"))
}
```

**v4:**

```go
func TestBlah(t *testing.T) {
    c := dial()

    var into interface{}
    err := c.Do(context.Background(), Cmd(&into, "UNKNOWN", "COMMAND"))

    var respErr resp3.SimpleError
    ok := errors.As(err, &respErr)

    assert.True(t, ok)
    assert.True(t, strings.HasPrefix(respErr.S, "ERR unknown command `UNKNOWN`"))
}
```

[...]

PR #306, opened 2021-12-29
Don't treat errors in aggregates as Go errors

### LINKED ISSUE #305: Incorrect deserialisation of an array with errors
Recently I started to work with RedisTimeSeries which has a [`TS.MADD` command](https://oss.redis.com/redistimeseries/commands/#tsmadd). The specialty of this command is the returned array which can contain both `int64` and `error` because in case of an error it continues the execution. This is not handled correctly in radix. Instead of deserialising each value radix fails at the first error it encounters.

Error:

```
response returned from Conn: unmarshaling message off Conn: ERR TSDB: the key is not a TSDB key
```

Example:

```go
package main

import (
	"context"
	"github.com/mediocregopher/radix/v4"
	"log"
	"testing"
	"time"
)

func TestClient_MAdd(t *testing.T) {
	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
	defer cancel()
	client, _ := (radix.PoolConfig{}).New(ctx, "tcp", "localhost:6379")
	err := client.Do(ctx, radix.FlatCmd(nil, "TS.CREATE", "known:key"))
	if err != nil {
		t.Error(err)
	}
	defer client.Do(ctx, radix.FlatCmd(nil, "DEL", "known:key"))
	var val interface{}
	args := []interface{}{
		"known:key", 946684800000, 1,
		"missing:key", 946684800000, 1,
		"known:key", 946684800001, 2,
	}
	err = client.Do(ctx, radix.FlatCmd(&val, "TS.MADD", args...))
	if err != nil {
		t.Error(err)
	}
	log.Println(val)
}
```

I tested the same command with `github.com/go-redis/redis/v8`, `github.com/gomodule/redigo`, and `github.com/joomcode/redispipe`. All three of them deserialises the array correctly (`int64`, `error`, `int64`).

nussjustin:
Possible solution for #305

Fixes #305

### DIFF
--- a/resp/resp.go
+++ b/resp/resp.go
@@ -51,9 +51,9 @@ type Opts struct {
 	// are emitted in a deterministic order.
 	Deterministic bool
 
-	// TreatErrorsAsValues indicates that unmarshaled RESP errors should
+	// DisableErrorBubbling indicates that unmarshaled RESP errors should
 	// not be treated as actual errors but like other response type.
-	TreatErrorsAsValues bool
+	DisableErrorBubbling bool
 }
 
 const defaultBytePoolThreshold = 10000000 // ~10MB
--- a/resp/resp.go
+++ b/resp/resp.go
@@ -50,6 +50,10 @@ type Opts struct {
 	// deterministic results. This is largely used for ensuring map key/values
 	// are emitted in a deterministic order.
 	Deterministic bool
+
+	// TreatErrorsAsValues indicates that unmarshaled RESP errors should
+	// not be treated as actual errors but like other response type.
+	TreatErrorsAsValues bool
 }
 
 const defaultBytePoolThreshold = 10000000 // ~10MB
--- a/resp/resp3/resp.go
+++ b/resp/resp3/resp.go
@@ -1677,7 +1677,7 @@ func Unmarshal(br resp.BufferedReader, rcv interface{}, o *resp.Opts) error {
 	}
 	prefix := Prefix(b[0])
 
-	if o == nil || !o.TreatErrorsAsValues {
+	if !o.TreatErrorsAsValues {
 		// if the prefix is one of the error types then just parse and return that
 		// full message here using the actual unmarshalers, which is easier than
 		// re-implementing them.
@@ -2009,7 +2009,7 @@ func unmarshalAgg(prefix Prefix, br resp.BufferedReader, l int64, rcv interface{
 		l *= 2
 	}
 
-	if o == nil || o.TreatErrorsAsValues == false {
+	if !o.TreatErrorsAsValues {
 		if o == nil {
 			o = resp.NewOpts()
 		} else {
--- a/resp/resp3/resp.go
+++ b/resp/resp3/resp.go
@@ -1609,6 +1609,8 @@ func saneDefault(prefix Prefix) (interface{}, error) {
 	// we don't handle ErrorPrefix because that always returns an error and
 	// doesn't touch I
 	switch prefix {
+	case BlobErrorPrefix, SimpleErrorPrefix:
+		return new(error), nil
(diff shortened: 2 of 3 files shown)


resp/resp3/resp.go
```diff
@@ -2004,6 +2011,17 @@ func unmarshalAgg(prefix Prefix, br resp.BufferedReader, l int64, rcv interface{
 		l *= 2
 	}
 
+	if o == nil || o.TreatErrorsAsValues == false {
```
mediocregopher: I don't think opts should ever be nil, so need to check for that.

resp/resp3/resp.go
```diff
@@ -1677,27 +1679,29 @@ func Unmarshal(br resp.BufferedReader, rcv interface{}, o *resp.Opts) error {
...
-	case BlobErrorPrefix:
-		var into BlobError
-		if err := into.UnmarshalRESP(br, o); err != nil {
-			return err
-		}
-		return resp.ErrConnUsable{Err: into}
-	case AttributeHeaderPrefix:
-		if err := DiscardAttribute(br, o); err != nil {
-			return err
+	if o == nil || !o.TreatErrorsAsValues {
```
mediocregopher: Again, opts should never be nil (I don't think? you've looked at this code more recently though lol)

mediocregopher (commented):
:pray:

resp/resp3/resp.go
```diff
@@ -1609,6 +1609,8 @@ func saneDefault(prefix Prefix) (interface{}, error) {
 	// we don't handle ErrorPrefix because that always returns an error and
 	// doesn't touch I
```
nerg4l: This comment might need some addition as errors will be handled here in case of `TreatErrorsAsValues`.

resp/resp3/resp.go
nussjustin: I just assumed it was ok for it to be nil, but it looks like all the code assumes it's non-nil, so I removed the check.

resp/resp3/resp.go
nussjustin: Done

resp/resp3/resp.go
nussjustin: The comment doesn't make much sense with the new code so I removed it.

resp/resp.go
```diff
@@ -50,6 +50,10 @@ type Opts struct {
 	// deterministic results. This is largely used for ensuring map key/values
 	// are emitted in a deterministic order.
 	Deterministic bool
+
+	// TreatErrorsAsValues indicates that unmarshaled RESP errors should
+	// not be treated as actual errors but like other response type.
+	TreatErrorsAsValues bool
```
nussjustin: Not really happy with the name, but all the others I came up with were even worse.

nussjustin:
This may actually also fix #296 

Not too sure about it, but sounds like it should.

resp/resp.go
nerg4l: What do you think about `DisableErrorBubbling`?

nussjustin: Ok, let's go with this.

mediocregopher:
This is looking good to me, @nussjustin do you think it's good to go?

nussjustin:
> This is looking good to me, @nussjustin do you think it's good to go?

Yes

nerg4l:
One last detail but this is something which can be handled in user space. If the returned value is not complex and can be casted to an error then it could make sense to return it as an error instead of a value.

mediocregopher:
> One last detail but this is something which can be handled in user space. If the returned value is not complex and can be casted to an error then it could make sense to return it as an error instead of a value.

I think this would unnecessarily complicate the behavior of the flag, especially since (as you said) it could be done in user-space.

PR #317, opened 2022-05-30
add a method getting the masterNode without lock to fix the deadlock for DoSecondary

dudusmida:
…for DoSecondary

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -500,10 +500,23 @@ func (c *Cluster) addrForKey(key string) string {
 	return ""
 }
 
+// add the getting master node without lock to fix the fix deadlock
+func (c *Cluster) addrForKeyWithNoLock(key string) string {
+	s := ClusterSlot([]byte(key))
+	for _, t := range c.primTopo {
+		for _, slot := range t.Slots {
+			if s >= slot[0] && s < slot[1] {
+				return t.Addr
+			}
+		}
+	}
+	return ""
+}
+
 func (c *Cluster) secondaryAddrForKey(key string) string {
 	c.l.RLock()
 	defer c.l.RUnlock()
-	primAddr := c.addrForKey(key)
+	primAddr := c.addrForKeyWithNoLock(key)
 	for addr := range c.secondaries[primAddr] {
 		return addr
 	}
--- a/cluster.go
+++ b/cluster.go
@@ -500,7 +500,7 @@ func (c *Cluster) addrForKey(key string) string {
 	return ""
 }
 
-// add the getting master node without lock to fix the fix deadlock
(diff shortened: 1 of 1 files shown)


mediocregopher:
Hi @dudusmida , thanks for the fix here! I'm not sure how this bug has made it this long, seems like something someone would have run into long ago.

Could you make the existing `addrForKey` method also use your `addrForKeyWithNoLock` method? That way we can avoid some duplicate code. Other than that this fix looks good to me.

nussjustin:
> I'm not sure how this bug has made it this long, seems like something someone would have run into long ago.

Both `secondaryAddrForKey` and `addrForKey` use a read lock so normally this is not a problem, except when some other goroutines tries to get a write lock after `secondaryAddrForKey` has obtained the read lock and before `addrForKey` tries to get the read lock again.

This is even [documented](https://pkg.go.dev/sync#RWMutex):

 > If a goroutine holds a RWMutex for reading and another goroutine might call Lock, no goroutine should expect to be able to acquire a read lock until the initial read lock is released. In particular, this prohibits recursive read locking. This is to ensure that the lock eventually becomes available; a blocked Lock call excludes new readers from acquiring the lock.

Guess someone _cough_ forgot this when writing the code.

dudusmida:
hi @mediocregopher , secondaryAddrForKey and addrForKey has used read lock . And in the secondaryAddrForKey，it use addrForKey to get the primary address. I think it make  nested lock. as @nussjustin said, when other goroutines tries to get a write lock ,it maby dead lock .  

i notice that the  function
`func (c *Cluster) sync(p Client, silenceFlag bool) error `  has used same lock with `function secondaryAddrForKey ` 
, and it try to get the write lock to update the Primaries and SecondaryOfAddr.


```
var l lock 
routineA
1)l.RLock
1.5）
2)l.RLock
3)l.RUnlock
4)l.RUnlock

routineB
1)
1.5) l.WLock
2)
```

- when routineA(A) steps (1.5),

- and routineB(B) is getting the write-lock, then the **B-(1.5)WLock** will block for waitting the **A-(1)Rlock**  unlock. 

- and routineA steps （2）, the **A-(2)Rlock** will block for wait the **B-(1.5)Wlock** unlock 
- then  it make dead lock:  
-- **A-(2)RLock** is waitting for  **B-(1.5)WLock** ,
-- **B-(1.5)WLock** is waiting for **A-(1)RLock** , 
-- **A-(1)RLock**  is waiting  for **A-(2)RLock**

mediocregopher:
@dudusmida I went ahead and cherry-picked your commits onto the `v3` branch and made a small modification (to make `addrForKey` also use `addrForKeyNoLock`). Thanks for the fix!

PR #323, opened 2022-07-01
Fix discarding of error values

### LINKED ISSUE #313: Unmarshaling problems
I am experiencing some unmarshaling errors with Radix when using Redis stream commands. I get a few different signatures:
invalid xread[group] response
expected prefix \"array\", got \"bulk-string\"
can't unmarshal into *[][]string, message body was: \"PONG\"

I think the root cause is that the Redis server have too low cpu and memory limits. In our production test environment (where the above is seen) k8s is used to deploy the Redis server and its limits are set to:
    limits:
      cpu: 50m
      memory: 70Mi
    requests:
      cpu: 30m
      memory: 30Mi
I also see a lot of:
read tcp [ipadress] i/o timeout
(Which also indicates that cpu and memory limits are too low).
In my local test environment where Redis server is deployed using docker without any limits I have never observed these issues.

But even if Redis is misbehaving (due too low cpu limits or other reasons) shouldnt Radix handle the situation. It seems like once the problem starts (Radix and Redis gets "out of sync"?) it seems like it never recovers. If the situation would recover once Redis server gets some "breathing space" I would be fine with it (not much Radix can do about a message from Redis it can not unmarshal other than returning an Error), but that doesnt seem to be the case. I think that Radix should be able to handle this situation better, i.e. recover.

Also is there something that can be done from the application to improve the situation? Configure longer timeouts in Radix? Reconnect to redis when the problem starts? Any other suggestions? 

Maybe this issue is somewhat related to issue #126 (but with different root cause) and #116

woodsbury:
Fixes #313 

When discarding the responses from previous commands that weren't
successfully read (e.g. if their context had a deadline that was
exceeded), RESP error values get treated the same as connection errors
and cause any reading from the connection to stop. This can cause a
response to not be read off of the connection or marked for discarding,
and therefore leave the connection out of sync.

This fix disables error bubbling during the discarding process to
prevent these error values from interfering.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -174,7 +174,12 @@ func (c *conn) reader(ctx context.Context) {
 					discardMU := el.Value.(connMarshalerUnmarshaler)
 
 					if err = resp3.Unmarshal(c.br, discardMU.unmarshalInto, c.rOpts); err != nil {
-						break
+						// Ignore RESP errors.
+						if !errors.As(err, &resp3.SimpleError{}) && !errors.As(err, &resp3.BlobError{}) {
+							break
+						}
+
+						err = nil
 					}
 
 					discardList.Remove(el)
--- a/conn_test.go
+++ b/conn_test.go
@@ -268,6 +268,96 @@ func TestConnDeadlineExceeded(t *T) {
 		t.Logf("successes:%d timeouts:%d closed:%d", numSuccesses, numTimeouts, numClosed)
 	})
 
+	t.Run("error", func(t *T) {
+		const pe, ps, n = 5, 5, 100
+		const initTimeout = time.Second
+		conn := dial()
+		defer conn.Close()
+		ctx := testCtx(t)
+
+		var wg sync.WaitGroup
+		wg.Add(pe + ps)
+
+		var numTimeouts, numClosed uint64
+
+		for i := 0; i < pe; i++ {
+			go func() {
+				timeout := initTimeout
+				defer wg.Done()
+				for i := 0; i < n; i++ {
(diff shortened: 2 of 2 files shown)


mediocregopher:
Hi @woodsbury , thanks for the PR! The solution you've given here was used as part of `unmarshalAgg`, but I don't think it is applicable to this case. It happens to work for your test case because you used strings as your `into` variables, and so when this line:

```
err = resp3.Unmarshal(c.br, discardMU.unmarshalInto, o)
```
is called it happens to be able to unmarshal the SimpleError into the string (because they can be coerced). But if you make this change:

```
diff --git a/conn_test.go b/conn_test.go
index 73abce0..b65318c 100644
--- a/conn_test.go
+++ b/conn_test.go
@@ -290,7 +290,7 @@ func TestConnDeadlineExceeded(t *T) {
                                        }

                                        innerCtx, cancel := context.WithTimeout(ctx, mkTimeout(timeout))
-                                       into := ""
+                                       into := 0
                                        err := conn.Do(innerCtx, Cmd(&into, "EVAL", "return redis.error_reply('NOTOK')", "0"))
                                        cancel()
```

you'll see the test breaks again.

You're original solution that you posted in the issue, however, works fine even with that `into := 0`, and I think is the right way to solve this problem. If you could use that solution in this PR instead I think we can merge this.

Thanks again for debugging this and putting together a test case, that's made reviewing this _much_ easier :pray:

woodsbury:
No problem, I've updated my changes to check the error type instead.

mediocregopher:
Looks good, thanks!

PR #325, opened 2022-08-01
Add special key handling for MEMORY USAGE

### LINKED ISSUE #324: MEMORY USAGE keys handling doesn't appear to be correct
There are currently a few special cases in actions.go for commands where the key isn't just the first argument, such as BITOP. One command that doesn't have any special handling at the moment, but perhaps should, is the MEMORY command which has a USAGE subcommand in it that takes a key.

At the moment sending a command like radix.Cmd(&value, "MEMORY", "USAGE", key) to a cluster seems to cause Radix to go to whichever shard the string "USAGE" is associated with, instead of the shard for the key.

woodsbury:
Add special handling for the MEMORY USAGE command to extract the key
from the second argument.

Fixes #324.


mediocregopher:
LGTM, thanks!

Issue #328, opened 2022-09-16
Race condition when refreshing pubSubConn


D-MarcFrank:
First of all: Thanks for sharing your great work!

The race condition occured when all Sentinel- and Redis-Nodes get shut down.

c.subs may be nil when called from persistentPubSub.refresh() where newPubSub() -> go c.spin() detected an error and closes the connection.

Instead of setting c.subs and c.psubs to nil in closeInner(), it would be better to fill them with an empty chanSet.

```
panic: assignment to entry in nil map

goroutine 26 [running]:
panic({0x86c300, 0x9b11e0})
        /usr/local/go/src/runtime/panic.go:987 +0x3ba fp=0xc0005317e8 sp=0xc000531728 pc=0x43837a
runtime.mapassign_faststr(0xc0008c6500?, 0x1?, {0x8e8d36, 0xd})
        /usr/local/go/src/runtime/map_faststr.go:205 +0x3da fp=0xc000531858 sp=0xc0005317e8 pc=0x4148da
github.com/mediocregopher/radix/v3.chanSet.add(...)
        ~/go/pkg/mod/github.com/mediocregopher/radix/v3@v3.7.0/pubsub.go:136
github.com/mediocregopher/radix/v3.(*pubSubConn).Subscribe(0xc0008c6500, 0x0?, {0xc000b3a3a0, 0x1, 0x4a56bc?})
        ~/go/pkg/mod/github.com/mediocregopher/radix/v3@v3.7.0/pubsub.go:387 +0x32f fp=0xc000531950 sp=0xc000531858 pc=0x6689ef
github.com/mediocregopher/radix/v3.(*persistentPubSub).refresh.func1()
        ~/go/pkg/mod/github.com/mediocregopher/radix/v3@v3.7.0/pubsub_persistent.go:157 +0x275 fp=0xc000531dc8 sp=0xc000531950 pc=0x66a575
github.com/mediocregopher/radix/v3.(*persistentPubSub).refresh(0xc0000ae080)
        ~/go/pkg/mod/github.com/mediocregopher/radix/v3@v3.7.0/pubsub_persistent.go:175 +0xbf fp=0xc000531e00 sp=0xc000531dc8 pc=0x66a25f
github.com/mediocregopher/radix/v3.(*persistentPubSub).spin(0xc0000ae080)
[...]

PR #329, opened 2022-09-16
Fix for #328: do not set chanSets to nil when closing

D-MarcFrank:


mediocregopher:
Thank you @D-MarcFrank ! The v3 pubsub code continues to prove to me that library packages should not deal in channels if they can help it :sweat_smile: . Thank you for the fix.

PR #350, opened 2023-11-29
Fix blocking when unmarshaling the slice of stream entries

pvragov:
Hello! There is a problem with unmarshaling the slice of StreamEntries when using the RESP3 protocol. In RESP3, the reply for XREADGROUP is a map. When Radix unmarshals this map, it doubles the expected size:
```
func unmarshalAgg(prefix Prefix, br resp.BufferedReader, l int64, rcv interface{}, o *resp.Opts) error {
	if prefix == MapHeaderPrefix {
		l *= 2
	}
        .....
        size := int(l)
```
This causes a block when attempting to use a slice of structs ([]StreamEntries) because the code below tries to fetch a doubled amount of elements:
```
		for i := 0; i < size || stream; i++ {
			into.Elem().Set(reflect.Zero(into.Type().Elem()))
			if more, err := maybeUnmarshalRESP(br, stream, into.Interface(), o); err != nil {
				return discardMultiAfterErr(br, size-i-1, err, o)
			} else if !more {
				break
			}
			slice = reflect.Append(slice, into.Elem())
		}
```

### DIFF
--- a/resp/resp3/resp.go
+++ b/resp/resp3/resp.go
@@ -2005,10 +2005,6 @@ func keyableReceiver(prefix Prefix, kv reflect.Value) (reflect.Value, error) {
 }
 
 func unmarshalAgg(prefix Prefix, br resp.BufferedReader, l int64, rcv interface{}, o *resp.Opts) error {
-	if prefix == MapHeaderPrefix {
-		l *= 2
-	}
-
 	if !o.DisableErrorBubbling {
 		if o == nil {
 			o = resp.NewOpts()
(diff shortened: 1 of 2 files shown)


mediocregopher:
Thank you for digging in and figuring this out, I'm sure all that reflection code was not fun to sort through :sweat_smile: I'm happy to merge this, but before I do would you mind adding a test which covers this case? Something which would have broken before this change. It doesn't have to be added to the `TestAnyUnmarshalMarshal` case, it can be a one-off like `TestMapIntoSliceOfStructs` or something like that. Just something to be sure this doesn't accidentally get broken in the future.

pvragov:
> Thank you for digging in and figuring this out, I'm sure all that reflection code was not fun to sort through 😅 I'm happy to merge this, but before I do would you mind adding a test which covers this case? Something which would have broken before this change. It doesn't have to be added to the `TestAnyUnmarshalMarshal` case, it can be a one-off like `TestMapIntoSliceOfStructs` or something like that. Just something to be sure this doesn't accidentally get broken in the future.

Hello. Thanks for the quick review. I've added the test you asked.

